PR #48, opened 2018-01-08
describe the most important configuration for the route, the method

xmonader:
the documentation is bit misleading when it comes to specify the method on the route, i had to do some hacks around it like dispatching manually based on the request method, and I was surprised from the tests that a `method` keyword is allowed in route function.


bigbes:
Rebased, applied to master branch. Thank you for your contribution!

PR #64, opened 2018-09-13
Update README.md

lav45:
Invalid method name https://github.com/tarantool/http/blob/master/http/server.lua#L269

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,8 +305,9 @@ function user_login(self)
 
     local user = box.select(users, 1, login, password)
     if user ~= nil then
-        return self:redirect_to('/'):
-            setcookie({ name = 'uid', value = user[0], expires = '+1y' })
+        local resp = self:redirect_to('/')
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -306,7 +306,7 @@ function user_login(self)
     local user = box.select(users, 1, login, password)
     if user ~= nil then
         return self:redirect_to('/'):
-            set_cookie({ name = 'uid', value = user[0], expires = '+1y' })
+            setcookie({ name = 'uid', value = user[0], expires = '+1y' })
```
Totktonada: 1. The description above lie about we should use `self:cookie(cookies)` (since 2fb0d0e91090efdc653063cd68ec57053c5bba14).
2. This should be before `return`.
3. Colon at the line above.
4. It seems it is the method of a response object and not a request object.

So this seems to be broken in more ways then fixed by this PR and the entire section need to be verified.

Totktonada:
Fixed in da7121cedf2614eebcd39cdc6a0c48ef75fde96b.

Issue #54, opened 2018-06-14
Debian package build for ubuntu:bionic fails


SudoBobo:
https://travis-ci.org/tarantool/http/jobs/391689680

PR #70, opened 2018-11-26
Add Ubuntu Bionic and Cosmic

Totktonada:


Totktonada:
Changed PACKAGECLOUD_REPO to 1_10, see #54.

PR #77, opened 2019-04-18
logging: impl custom log option per route

artur-barsegyan:


sharonovd:
Generally LGTM for me

test/http.test.lua
```diff
@@ -388,4 +413,81 @@ test:test("server requests", function(test)
...
+
+        httpd:stop()
+    end)
+
+    test:test("Log route requests with turned off 'log_requests' option", function(test)
+        test:plan(1)
+        local httpd_without_logs = http_server.new("127.0.0.1", 12345, { log_requests = false, log_errors = false })
+        httpd_without_logs:start()
+
+        httpd:route({ path = '/', log_level = 3 }, function(_) end)
```
ASverdlov: `httpd_without_logs:route` maybe?
After this fix the test should fail.

README.md
```diff
@@ -158,6 +158,10 @@ The first argument for `route()` is a Lua table with one or more keys:
 * `path` - route path, as described earlier.
 * `name` - route name.
 * `method` - method on the route like `POST`, `GET`, `PUT`, `DELETE`
+* `log_level` - type of logger specified for logging request info.
+  Any digit from range 1 - 6 is allowed, where: `1` is disable logging - `6` is `log.verbose`.
+  By default, `log_level` is equal to 5 (`log.info`). If `log_requests` option is `false` this parameter will be ignored.
```
ASverdlov: > If `log_requests` option is `false` this parameter will be ignored. 
This is not true. According to your code, `log_level` just rewrites a log function and doesn't depend on `log_requests` value anyhow.

test/http.test.lua
artur-barsegyan: yeah
you right. copy - paste is evil

http/server.lua
```diff
@@ -589,7 +589,23 @@ response_mt = {
...
+    return log.info
+end
+
+local function set_error_logger(route)
+    if route and route.endpoint.err_logger ~= nil then
+        return route.endpoint.err_logger
+    end
+
+    return log.error
+end
```
filonenko-mikhail: функции больше похожи на get_*** чем на set_***

artur-barsegyan: in english, please!
soglasen Mish

ASverdlov:
We have another feature that is really needed: custom logger.
If we accept this feature, it won't solve it and will introduce unneeded complexity in the future.

I propose adding new route options (which we have in server already): `log_errors `, `log_requests`. Extend their semantic in both server and route: accept `function` type (with log.info semantics) and `boolean` type (legacy behaviour).

In case route option is specified - it overrides the server one.

ASverdlov:
As discussed privately, LuaJit errors on sockets introduced by new tests need to be fixed.

artur-barsegyan:
> As discussed privately, LuaJit errors on sockets introduced by new tests need to be fixed.

I post a new comment to [this bug issue ](https://github.com/tarantool/tarantool/issues/4087). Actually, it's tarantool socket implementation bug

README.md
```diff
@@ -110,8 +110,13 @@ httpd = require('http.server').new(host, port[, { options } ])
   type `text/html`, `text/plain` and `application/json`.
 * `display_errors` - return application errors and backtraces to the client
   (like PHP).
-* `log_errors` - log application errors using `log.error()`.
-* `log_requests` - log incoming requests.
+* `log_requests` - log incoming requests. This parameter can receive:
+    - custom log function using for all routes in current instance of http server. The function should receive a format string
```
ASverdlov: Faster to read version:

function value, supporting C-style formatting: `log_requests(fmt, ...)`, where `fmt` is a format string and `...` is Lua Varargs, holding arguments to be replaced in fmt.

README.md
```diff
@@ -110,8 +110,13 @@ httpd = require('http.server').new(host, port[, { options } ])
...
   (like PHP).
-* `log_errors` - log application errors using `log.error()`.
-* `log_requests` - log incoming requests.
+* `log_requests` - log incoming requests. This parameter can receive:
+    - custom log function using for all routes in current instance of http server. The function should receive a format string 
+      as the first argument and the following variable number of arguments uses with format string.
+    - boolean value, where `true` choose default `log.info` and `false` disable request logs at all. 
+  
+  By default uses `log.info` function for requests logging. 
+* `log_errors` - same as the `log_requests` option but uses for error messages logging. By default uses `log.error()` function.
```
ASverdlov: Grammar:
> but uses for
but is used for

http/server.lua
```diff
@@ -589,7 +589,53 @@ response_mt = {
     }
 }
 
-local function handler(self, request)
+local function is_function(obj)
+    return type(obj) == 'function'
+end
+
+local function get_request_logger(server_opts, route_opts)
+    local root_logger = log.debug
```
ASverdlov: Simplification suggestion:
1) check route first, if found - return immediately, then
2) check server, if found - return immediately, then
3) return default

No need for `root_logger` variable.

http/server.lua
```diff
@@ -589,7 +589,53 @@ response_mt = {
...
+        elseif route_opts.endpoint.log_requests == false then
+            return log.debug
+        end
+    end
+
+    return root_logger
+end
+
+local function get_error_logger(server_opts, route_opts)
+    local root_logger = log.debug
```
ASverdlov: Simplification suggestion:
1) check route first, if found - return immediately, then
2) check server, if found - return immediately, then
3) return default

No need for `root_logger` variable.

http/server.lua
```diff
@@ -685,21 +729,22 @@ local function process_client(self, s, peer)
             s:write('HTTP/1.0 100 Continue\r\n\r\n')
         end
 
-        local logreq = self.options.log_requests and log.info or log.debug
+        local route = self:match(p.method, p.path)
```
ASverdlov: Add comment describing what you did here!

Something like:
`-- HACK: we're in a server context, but prematurely dispatch route (router responsibility) for getting logging options which override server ones' `.

http/server.lua
```diff
@@ -685,21 +729,22 @@ local function process_client(self, s, peer)
...
+        local logreq = get_request_logger(self.options, route)
         logreq("%s %s%s", p.method, p.path,
-            p.query ~= "" and "?"..p.query or "")
+               p.query ~= "" and "?"..p.query or "")
 
-        local res, reason = pcall(self.options.handler, self, p)
+        local res, reason = pcall(self.options.handler, self, route, p)
```
ASverdlov: Maybe better leave double-dispatch and not change router signature?

http/server.lua
artur-barsegyan: I think this is obvious and no comment is needed.

http/server.lua
artur-barsegyan: Yes, it will be more obvious

ASverdlov:
LGTM.

PR #79, opened 2019-04-20
Check ptr to string before use it

### LINKED ISSUE #51: httpd: SEGV_MAPERR when box httpd html escaping
# Environment

```
tarantool --v
Tarantool 1.9.0-61-gcd48321
Target: Darwin-x86_64-Release
Build options: cmake . -DCMAKE_INSTALL_PREFIX=/usr/local/Cellar/tarantool/1.9.0.61 -DENABLE_BACKTRACE=ON
Compiler: /usr/local/Homebrew/Library/Homebrew/shims/super/clang /usr/local/Homebrew/Library/Homebrew/shims/super/clang++
C_FLAGS: -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -fopenmp -msse2 -std=c11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation
CXX_FLAGS: -fexceptions -funwind-tables -fno-omit-frame-pointer -fno-stack-protector -fno-common -fopenmp -msse2 -std=c++11 -Wall -Wextra -Wno-strict-aliasing -Wno-char-subscripts -Wno-format-truncation
```

# Test case

```
tarantoolctl rocks install html
```

test.lua 

```lua

!/usr/bin/env tarantool

require('strict').on()

local http = require('http.server')

local server = http.new("0.0.0.0",9090)

local test_el = [[
      <%= {{continue}} %>"
]]

local function test(req)

    local resp = req:render({continue="/"})
    resp.status = 200
    return resp
end


server:route({path='/', template=test_el}, test)
server:start()
```

send request

```
curl http://127.0.0.1:9090/
```

# Backtrace

```
Segmentation fault
  code: SEGV_MAPERR
  addr: 0x0
  context: 0x102c1f4c8
  siginfo: 0x102c1f460
Current time: 1523973121
Please file a bug at http://github.com/tarantool/tarantool/issues
Attempting backtrace... Note: since the server has already crashed,
this may fail as well
#0  0x10071e6be in print_backtrace+9
#1  0x10067cb03 in _ZL12sig_fatal_cbiP9__siginfoPv+158
#2  0x7fff7aeb7f5a in _sigtramp+1a
#3  0x102b14f82 in lbox_httpd_escape_html+d2
#4  0x100733005 in lj_BC_FUNCC+44
#5  0x100743003 in lua_pcall+96
#6  0x102b13a7d in lbox_httpd_template+4ad
#7  0x100733005 in lj_BC_FUNCC+44
#8  0x100743003 in lua_pcall+96
#9  0x10071164c in luaT_call+13
#10 0x10070dc8b in lua_fiber_run_f+66
#11 0x10067c6fc in _ZL16fiber_cxx_invokePFiP13__va_list_tagES0_+c
#12 0x10071c878 in fiber_loop+2a
#13 0x100831267 in coro_init+4b
abort trap: 6
```

# Other test case

```
#!/usr/bin/env tarantool

require('strict').on()

local http = require('http.server')

local server = http.new("0.0.0.0",9090)

local test_el = [[

      % if error then
        <%= error %>
      % end
]]

local function test(req)

    local resp = req:render({})
    resp.status = 200
    return resp
end


server:route({path='/', template=test_el}, test)
server:start()
```

olegrok:
Closes #51

### DIFF
--- a/http/lib.c
+++ b/http/lib.c
@@ -91,6 +91,37 @@ tpl_term(int type, const char *str, size_t len, void *data)
 	}
 }
 
+/**
+ * This function exists because lua_tostring does not use
+ * __tostring metamethod, and this metamethod has to be used
+ * if we want to print Lua userdata correctly.
+ */
+static const char *
+luaT_tolstring(lua_State *L, int idx, size_t *len)
+{
+	if (!luaL_callmeta(L, idx, "__tostring")) {
+		switch (lua_type(L, idx)) {
+			case LUA_TNUMBER:
+			case LUA_TSTRING:
+				lua_pushvalue(L, idx);
+				break;
+			case LUA_TBOOLEAN: {
+				int val = lua_toboolean(L, idx);
+				lua_pushstring(L, val ? "true" : "false");
+				break;
+			}
+			case LUA_TNIL:
+				lua_pushliteral(L, "nil");
+				break;
+			default:
+				lua_pushfstring(L, "%s: %p", luaL_typename(L, idx),
+								lua_topointer(L, idx));
+		}
+	}
+
+	return lua_tolstring(L, -1, len);
+}
+
 static int
 lbox_httpd_escape_html(struct lua_State *L)
 {
@@ -109,11 +140,8 @@ lbox_httpd_escape_html(struct lua_State *L)
 	}
 
 	for (i = 1; i <= top; i++) {
-		if (lua_isnil(L, i)) {
-			luaL_addstring(&b, "nil");
-			continue;
-		}
-		const char *s = lua_tostring(L, i);
+		const char *s = luaT_tolstring(L, i, NULL);
+		int str_idx = lua_gettop(L);
 		for (; *s; s++) {
 			switch(*s) {
 				case '&':
@@ -136,6 +164,7 @@ lbox_httpd_escape_html(struct lua_State *L)
 					break;
 			}
 		}
+		lua_remove(L, str_idx);
 	}
 
 	luaL_pushresult(&b);
--- a/test/unit/lib_test.lua
+++ b/test/unit/lib_test.lua
@@ -37,6 +37,11 @@ g.test_template = function()
     local rendered = http_lib.template(template, { t = tt })
     t.assert(#rendered > 10000, "rendered size")
     t.assert_equals(rendered:sub(#rendered - 7, #rendered - 1), "</html>", "rendered eof")
+
+    -- gh-51 Incorrect arguments escaping leads to segfault
+    template = [[<%= {{continue}} %>"]]
+    local result = http_lib.template(template, {continue = '/'})
+    t.assert(result:find('\"') ~= nil)
 end
 
 g.test_parse_request = function()


http/lib.c
```diff
@@ -114,6 +114,9 @@ lbox_httpd_escape_html(struct lua_State *L)
 			continue;
 		}
 		const char *s = lua_tostring(L, i);
+		if (s == NULL) {
```
Totktonada: Thanks for the investigation.

I think the fix is not quite correct. We should do something like tostring() for a returned value, not just ignore it. Please, consider copying of luaT_tolstring() into lib.c source code.

https://github.com/tarantool/tarantool/commit/01c3270117b337cb8f1c0c1309d3ae7f46f9a948

olegrok: This function lead to CI fail (and it's quite strange) - I've adopted it

http/lib.c
```diff
@@ -91,6 +91,36 @@ tpl_term(int type, const char *str, size_t len, void *data)
...
+			case LUA_TNIL:
+				lua_pushliteral(L, "nil");
+				break;
+			default:
+				lua_pushfstring(L, "%s: %p", luaL_typename(L, idx),
+								lua_topointer(L, idx));
+		}
+	}
+
+	return lua_tostring(L, -1);
```
Totktonada, line 121: `s/-1/idx/`

(I see, the same mistake exists in tarantool code.)

Totktonada, line 121: The function should leave the Lua stack top unchanged.

(And, yep, the same problem exists in the tarantool code.)

http/lib.c
Totktonada: Maybe due to changing of the Lua stack top. Anyway, I left a comment regarding this, so this one can be resolved.

Totktonada (approved):
The fix looks good to me, except a couple of details above.

I would also ask to add a test if the module would be in my responsibility.

CC @Satbek.

http/lib.c
Totktonada: My bad, the new value is always pushed at top of the stack. Sorry.

olegrok: I understood a root of problem and just added lua_pop after function call. Now this module contains complete copy of luaT_tolstring function from "module.h". It allows to preserve compatibility with older versions.

Also test was introduced.

Thanks for your review!

Totktonada: Related: https://github.com/tarantool/tarantool/commit/fa0f10cb0b4706c35f3438c980be3d414e5c55d6

I'm not sure about this for now, so not sent yet.

olegrok: I noticed strange behaviour (only on centos 7). Something was wrong with Lua stack, I can't just use lua_pop after luaL_addchar/add_string but after I changed lua_pop with lua_gettop + lua_remove tests were passed.

I think we could check your patch.

However, according Lua C API reference (https://pgl.yoyo.org/luai/i/lua_tolstring):
```
lua_tolstring returns a fully aligned pointer to a string inside the Lua state.
This string always has a zero ('\0') after its last character (as in C), but can contain other zeros in its body.
Because Lua has garbage collection, there is no guarantee that the pointer returned by lua_tolstring will be valid after the corresponding value is removed from the stack.
```

You can't just get a pointer to the top of string and then change a top.
If this string is garbage collected it will be a use-after-free:
```C
const char *res = lua_tolstring(L, -1, len); // <--- valid pointer
lua_settop(L, top);
// -- collectgarbage ---
return res; // -- invalid pointer
```

Totktonada: D'oh! Thanks for the notice.

PR #85, opened 2019-07-08
Replace req in before_dispatch and after_dispatch

norguhtar:
Motivation. We need this change for simple add auth checks and set headers. Also with this you can fully handle auth check and redirect in before_dispatch, and can use after_dispatch for set headers or change response on all routes.

What change and how it works added in doc


README.md
```diff
@@ -191,59 +191,64 @@ In that case, the handler body is taken from a file in the
...
 
-### Fields and methods of the Request object
+### Fields and methods of the Context object
 
-* `req.method` - HTTP request type (`GET`, `POST` etc).
-* `req.path` - request path.
-* `req.query` - request arguments.
-* `req.proto` - HTTP version (for example, `{ 1, 1 }` is `HTTP/1.1`).
-* `req.headers` - normalized request headers. A normalized header
+* `ctx.req.body` - request's body
```
ASverdlov: This is a breaking change.

It will never be merged in `tarantool/http` v1.

ASverdlov:
We don't allow breaking changes in the same major version.

For future: be nice - separate refactoring from the actual code, or better - put it in another PR.

README.md
norguhtar: Okay. We just use our version. Current tarantool/http v1 not usefull for us. Notify me when v2 go to beta.

PR #93, opened 2019-10-16
uri_escape default configuration

migolovanov:
By default URL-decode is not working for `+` sign in GET-arguments.
But since other web-servers perform such decoding, i believe it should be enabled.

### DIFF
--- a/http/server.lua
+++ b/http/server.lua
@@ -186,7 +186,7 @@ local function query_param(self, name)
             local params = lib.params(self.query)
             local pres = {}
             for k, v in pairs(params) do
-                pres[ uri_unescape(k) ] = uri_unescape(v)
+                pres[ uri_unescape(k, true) ] = uri_unescape(v, true)
             end
             rawset(self, 'query_params', pres)
         end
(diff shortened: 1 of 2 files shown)


Mons (approved):
LGTM

kyukhin:
Please, rebase your changes.

Totktonada:
The idea seems right, but:

1. The `master` goes forward, the patch conflicts with it now.
2. We should cover the change with a test.
3. We should unescape keys with the plus sign decoding, not only values.

@0x501D, can you proceed with those changes, please?

----

It is not the topic here, but I also looked at the POST parameters decoding:

https://github.com/tarantool/http/blob/13b0e2fcbc7a64e83fb234a72a366ad747bc2274/http/server.lua#L200-L225

1. In the `application/x-www-form-urlencoded` case we should also decode the plus sign in keys.
2. The `else` case is strange or I don't know something. It is like `application/x-www-form-urlencoded`, but without plus decoding. I don't know, what is the idea.

0x501D:
> @0x501D, can you proceed with those changes, please?

Done

Totktonada (approved):
Thank you for the update! The implementation seems reasonable for me. LGTM.

Nit: I would squash the commits into one.

PR #105, opened 2020-01-27
Fix routes overlapping by *any pattern in route's path

Satbek:
Closes #105 

The bug was in incorrect ``specificity`` counting. 
``specificity`` is a how many symbols were not known (were hidden behind : and * patterns) in route's path. ``specificity`` was calculated as negative sum of symbols in regex groups, which were created from route's path.

 ```lua
local regex_groups_matched = {string.match(filter.path, r.match)}
```

But if match pattern has not any regex groups and string matches,  string.match returns all string.

Example:

```lua
tarantool> string.match('test/some', '^test/(.-)$')
---
- some
...

tarantool> string.match('test/some', '^test/some$')
---
- test/some
...
```

So if request's path is the same as route's path then string.match will return whole and this route will get the lowest priority.

So we need to explicitly add regex group which covers all path and take this into account.

```lua
tarantool> string.match('test/some', '^(test/(.-))$')
---
- test/some
- some
...

### DIFF
--- a/http/router/matching.lua
+++ b/http/router/matching.lua
@@ -57,7 +57,7 @@ local function transform_pattern(path)
         if string.match(match, '^.') ~= '/' then
             match = '/' .. match
         end
-        match = '^' .. match .. '$'
+        match = '^(' .. match .. ')$'
     end
 
     return match, stash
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: http/router/matching.lua [hunk 1]
```diff
@@ -70,13 +70,20 @@ local function matches(r, filter)
...
+
     if #regex_groups_matched == 0 then
         return false
     end
-    if #r.stash > 0 and #r.stash ~= #regex_groups_matched then
+    if #r.stash > 0 and #r.stash ~= #regex_groups_matched - 1 then
         return false
     end
 
+    -- first regex_group is whole path, we should omit it
```
rosik: Why not simply `table.remove(regex_groups_matched, 1)`?

Satbek: Yes it's simpler and clearer.

PR #119, opened 2020-05-13
Get rid of io module.

### LINKED ISSUE #112: Use fio instead of io
Subj: https://github.com/tarantool/http/blob/master/http/router/fs.lua#L65

Satbek:
In tarantool the standart built-in module for working with files is "fio".
So we should use it instead of "io".

Closes #112

### DIFF
--- a/http/router/fs.lua
+++ b/http/router/fs.lua
@@ -4,6 +4,7 @@ local mime_types = require('http.mime_types')
 local response = require('http.router.response')
 
 local json = require('json')
+local fio = require('fio')
 
 local function type_by_format(fmt)
     if fmt == nil then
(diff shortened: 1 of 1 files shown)


http/router/fs.lua
```diff
@@ -62,14 +63,14 @@ local function static_file(self, request, format)
         }
     end
 
-    local s, fh = pcall(io.input, file)
+    local fh, _ = fio.open(file, {'O_RDONLY'})
```
olegrok, line 66: Why don't you handle possible error?

http/router/fs.lua
```diff
@@ -62,14 +63,14 @@ local function static_file(self, request, format)
...
 
-    if not s then
+    if fh == nil then
         return { status = 404 }
     end
 
-    local body = fh:read('*a')
-    io.close(fh)
+    local body = fh:read()
```
olegrok, line 72: I think we should check errors in all cases and handle them properly. Not just ignore.

http/router/fs.lua
```diff
@@ -62,14 +64,19 @@ local function static_file(self, request, format)
...
+    if err ~= nil then
         return { status = 404 }
     end
 
-    local body = fh:read('*a')
-    io.close(fh)
+    local body
+    body, err = fh:read()
+    if err ~= nil then
+        utils.errorf("Can not return static file for '%s': '%s'", request:path(), err)
```
olegrok, line 76: After we will return {status = 200, body = nil}. 200 - OK, but we have an error it's not OK.

Satbek: No. It's OK. After we will return error response with status = 500.

olegrok: Hm...

It's clear above when we `return { status = 404 }`, but completely unclear here

olegrok: But anyway it's better than it was before. I mean io to fio change

PR #122, opened 2020-05-14
Fixed middleware routing.

### LINKED ISSUE #121: middleware does not match '/' route
assertion will fail

```lua
local server = require('http.server').new(0, 8080)
local router = require('http.router').new()
router:use(function() return {body = 'middleware', status = 200} end, {})
server:set_router(router)
server:start()
local r = http_client.get('localhost:8080/')
assert(r.status == 200)
server:stop()
```

Satbek:
Now middleware matches '/' path.

Closes #121

### DIFF
--- a/http/router/init.lua
+++ b/http/router/init.lua
@@ -175,11 +175,12 @@ local function use_middleware(self, handler, opts)
         before = '?string|table',
         after = '?string|table',
     })
+    opts = opts or {}
     opts = table.deepcopy(opts)
     opts.handler = handler
 
     local uuid = require('uuid')
-    opts.path = opts.path or '/.*'
+    opts.path = opts.path or '?.*'
     opts.method = opts.method or 'ANY'
     opts.name = opts.name or uuid.str()
     opts.before = opts.before or {}
--- a/test/integration/middleware_test.lua
+++ b/test/integration/middleware_test.lua
@@ -1,0 +1,136 @@
+local t = require('luatest')
+local g = t.group()
+local helper = require('test.helper')
+local http_server = require('http.server')
+local http_router = require('http.router')
+
+local json = require('json')
+local http_client = require('http.client')
+local fio = require('fio')
+
+
+g.before_each(function()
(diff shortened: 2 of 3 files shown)


test/integration/middleware_test.lua
```diff
@@ -0,0 +1,137 @@
...
+        local lua_body = json.decode(resp.body)
+        lua_body.message = 'hello world!'
+        resp.body = json.encode(lua_body)
+
+        return resp
+    end
+
+    local ok = g.router:use(add_helloworld_to_response, {
+        name = 'hello_world',
+        path = '/.*',
```
printercu, line 46: Is this the same as omitting `path`? If so, let's omit it.

test/integration/middleware_test.lua
```diff
@@ -0,0 +1,137 @@
...
+        'http://127.0.0.1:12345/fruits/apple'
+    )
+    t.assert_equals(r.status, 200, 'status')
+    parsed_body = json.decode(r.body)
+    t.assert_equals(parsed_body.kind, 'orange', 'route swapped from apple handler to orange')
+end
+
+g.test_middleware_routing = function()
+    g.router:route({path = 'some_path'}, function() return {body = 'from_route', status = 200} end)
+    g.router:use(function() return {body = 'middleware', status = 200} end, {})
```
printercu, line 129: Let's make it possible to `:use(middleware)` without empty options.

http/router/init.lua
```diff
@@ -179,7 +179,7 @@ local function use_middleware(self, handler, opts)
     opts.handler = handler
 
     local uuid = require('uuid')
-    opts.path = opts.path or '/.*'
+    opts.path = opts.path or '?.*'
```
printercu, line 182: what is `?` for? Doesn't it work with `.*` or empty string?

Totktonada:
We finally discarded http v2: https://github.com/tarantool/http/releases/tag/1.1.1.

PR #123, opened 2020-05-28
Don't append slash to path to make `/.*` pattern work for `/`

printercu:


ligurio:
PR is in draft for last 2 years. Feel free to reopen.

PR #126, opened 2020-06-25
Added ability to set cookie without escaping.

### LINKED ISSUE #114: Fix not necessary url_escape in cookie
It already fixed in v1. but this bug still exist in v2. Can you fixed this?

https://github.com/tarantool/http/pull/82

Satbek:
`resp:setcookie` implicitly escaped cookie values. Added ability to set cookie without any escaping `resp:setcookie('name', 'value', {raw = true})`.
Also added escaping for cookie path, and changed escaping algorithm according to https://tools.ietf.org/html/rfc6265.

Closes #114

### DIFF
--- a/http/router/response.lua
+++ b/http/router/response.lua
@@ -28,7 +28,43 @@ local function expires_str(str)
     return os.date(fmt, gmtnow + diff)
 end
 
-local function setcookie(resp, cookie)
+local function valid_cookie_value_byte(byte)
+    -- https://tools.ietf.org/html/rfc6265#section-4.1.1
+    -- US-ASCII characters excluding CTLs, whitespace DQUOTE, comma, semicolon, and backslash
+    return 32 < byte and byte < 127 and byte ~= string.byte('"') and
+            byte ~= string.byte(",") and byte ~= string.byte(";") and byte ~= string.byte("\\")
+end
+
+local function valid_cookie_path_byte(byte)
+    -- https://tools.ietf.org/html/rfc6265#section-4.1.1
+    -- <any CHAR except CTLs or ";">
+    return 32 <= byte and byte < 127 and byte ~= string.byte(";")
+end
+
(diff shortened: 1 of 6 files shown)


http/utils.lua
```diff
@@ -36,6 +36,14 @@ local function extend(tbl, tblu, raise)
     return res
 end
 
+local function base_escape(char)
+    return string.format('%%%02X', string.byte(char))
+end
+
+local function base_unescape(char)
```
printercu, line 43: is it base16_decode/encode?

http/router/response.lua
```diff
@@ -28,7 +28,46 @@ local function expires_str(str)
...
+
+local function escape_path(cookie_path)
+    return escape_string(cookie_path, valid_cookie_path_byte)
+end
+
+local function setcookie(resp, cookie, options)
+    if options == nil then
+        options = {}
+    end
+    options = utils.extend({raw = false}, options, true)
```
printercu, line 69: I think this line is unnecessary, because `if not {}.raw` is the same as `if not {raw = false}.raw`.

what about `options = options or {}`

test/unit/setcookie_test.lua
```diff
@@ -0,0 +1,176 @@
...
+                    value = "f" .. string.char(byte) .. "f",
+                    result = "f" .. string.char(byte) .. "f",
+                }
+        end
+    end
+
+    for case_name, case in pairs(test_table) do
+        local resp = get_object()
+        resp:setcookie({ name='name', value = case.value })
+        t.assert_equals(resp.headers['set-cookie'], {"name=" .. case.result}, case_name)
```
printercu, line 55: just a tip. such loops can be written in this way
```
local function assert_escaped_value(input, expected)
...
end
assert_escaped_value("f f", 'f%20f')
assert_escaped_value('f"f', 'f%22f')
```

PR #132, opened 2021-02-03
Added compatibility with v1 API.

### LINKED ISSUE #131: Backward compatibility with http-v1 api
Let's add ability to use [v1 API ](https://github.com/tarantool/http/tree/1.1.0). 

Due to incompatibility of http-server versions v1 API would work via adapter layer, like [here](https://github.com/tarantool/cartridge/pull/526/files).

The incompatibility list:
- splitted router and server objects in http v2, in http v1 they are parts of one object
- before_dispatch and after_dispatch hooks are not supported by default
- headers, path, peer, method, proto, query in http v1 are request's object fields, but they are methods in http v2
- ``handler`` option was removed from v2

Also user should get a warning, if he uses v1 API. Both versions' usage should be prohibited, user should use only [v2 API ](https://github.com/tarantool/http/tree/master) or [v1 API](https://github.com/tarantool/http/tree/1.1.0).

Satbek:
It supports all v1 methods, options, fields. It works via layer over v2 API.
But user can use only one API. http_server will throw an error if user tries to use methods from both APIs.

closes #131

### DIFF
--- a/http/server.lua
+++ b/http/server.lua
@@ -1,0 +1,297 @@
+local v2_server = require('http.server.init')
+local v2_router = require('http.router')
+local v1_server_adapter = require('http.v1_server_adapter')
+local log = require('log')
+local API_VERSIONS = require('http.api_versions')
+
+local function httpd_stop(self)
+    return self.__v2_server:stop()
+end
+
+local function httpd_set_router(self, router)
+    return self.__v2_server:set_router(router)
+end
+
+local function httpd_router(self)
+    return self.__v2_server:router()
+end
+
+local function __set_v1_handler(self, handler)
+    return self.__v1_server_adapter:set_handler(handler)
+end
+
+local function server_route(self, opts, handler)
+    self.__v1_server_adapter:route(opts, handler)
+    return self
+end
+
+local function server_match(self, method, route)
+    return self.__v1_server_adapter:match(method, route)
+end
+
+local function server_helper(self, name, handler)
+    self.__v1_server_adapter:helper(name, handler)
+    return self
+end
+
+local function server_hook(self, name, handler)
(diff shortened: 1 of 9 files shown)


http/server.lua
```diff
@@ -0,0 +1,297 @@
...
+
+local function v1_method_decorator(method, method_name)
+    return function(self, ...)
+        if self.__api_version == API_VERSIONS.V1 then
+            return method(self, ...)
+        elseif self.__api_version == API_VERSIONS.V2 then
+            error(
+                ('":%s" method does not supported. Use http-v2 api https://github.com/tarantool/http/tree/master.'):
+                format(method_name)
+            )
```
Totktonada, line 119: Is the check http server instance wide? So, different http server instances may be used according to different API versions, right? Is it fit good to the case, when cartridge uses one API version, but a user's application uses another one?

What is the reason of blocking such usage? It would not work appropriately?

Satbek: It is a check only for one http_server instance, not for all.
Yes, different instances could be used according to different API versions.
Yes, I think it fits good to the case with user and cartridge application servers. We may ask @rosik about it.

Both APIs have documentation and covered by tests. But their simultaneous usage not documented and not tests. So I think there are cases when simultaneous usage can lead to bugs or strange not documented behaviour. Error on such cases prevents it. Also I don't think that simultaneous usage is good practice. As for me it would not give any advantages and confuses a code with a lot of method calls, which do the same thing.

Totktonada: Thanks for the clarification! It seems meaningful.

Totktonada: @rosik said that users usually add application specific routes into the cartridge's http server.

Satbek:
no need on it after https://github.com/tarantool/http/issues/134

PR #141, opened 2021-10-20
Add options to control keepalive connections

### LINKED ISSUE #137: Handle idle connections
### Problem statement

We have http server exposing metrics to telegraf. Telegraf scrapes metrics in 60 second intervals via http: it opens keep-alive connections but under certain conditions (have no repro, only in prod) it does not reuses established connections and opens new connection each time it scrapes metrics. This results in increasing number of open connections, every connection if a fiber which lives forever. 

### Possible solution

Introduce new http.server config options:
- maximum number of idle connections, after reaching this limit http.server could start closing idle conns
- idle timeout is the maximum amount of time an idle (keep-alive) connection will remain idle before closing

ligurio:
Fixes #137.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,9 @@ httpd = require('http.server').new(host, port[, { options } ])
   })
   ```
 
+* `idle_timeout` - maximum amount of time an idle (keep-alive) connection will
+  remain idle before closing. When the idle timeout is exceeded, HTTP server
+  closes the keepalive connection. Default value: 0 seconds (disabled).
 
 ## Using routes
 
--- a/README.md
+++ b/README.md
@@ -132,6 +132,20 @@ httpd = require('http.server').new(host, port[, { options } ])
 
   By default uses `log.info` function for requests logging.
 * `log_errors` - same as the `log_requests` option but is used for error messages logging. By default uses `log.error()` function.
+* `disable_keepalive` - disables keep-alive connections with misbehaving
+  clients. Parameter accept a table that contains a user agents names.
+  By default table is empty.
+
+  Example:
+
+  ```lua
+  local httpd = http_server.new('127.0.0.1', 8080, {
+      log_requests = true,
+      log_errors = true,
+      disable_keepalive = { 'curl/7.68.0' }
+  })
+  ```
+
 
 ## Using routes
 
--- a/http/server.lua
+++ b/http/server.lua
@@ -23,6 +23,16 @@ local function sprintf(fmt, ...)
     return string.format(fmt, ...)
 end
 
+-- Converts a table to a map, values becomes keys with value 'true'.
+-- { 'a', 'b', 'c' } -> { 'a' = true, 'b' == 'true', 'c' = true }
+local function tomap(tbl)
+    local map = {}
+    for _, v in pairs(tbl) do
+        map[v] = true
+    end
+    return map
+end
+
 local function valid_cookie_value_byte(byte)
     -- https://tools.ietf.org/html/rfc6265#section-4.1.1
     -- US-ASCII characters excluding CTLs, whitespace DQUOTE, comma, semicolon,
@@ -891,6 +901,11 @@ local function process_client(self, s, peer)
             end
         end
 
+        local useragent = p.headers['user-agent']
+        if self.disable_keepalive[useragent] == true then
+            hdrs.connection = 'close'
+        end
+
         local response = {
             "HTTP/1.1 ";
             status;
@@ -1259,7 +1274,9 @@ local function httpd_start(self)
     local server = self.tcp_server_f(self.host, self.port, {
         name = 'http',
         handler = function(...)
+            self.internal.preprocess_client_handler()
             process_client(self, ...)
+            self.internal.postprocess_client_handler()
         end,
         http_server = self,
     })
@@ -1285,6 +1302,11 @@ local exports = {
         if type(options) ~= 'table' then
             errorf("options must be table not '%s'", type(options))
         end
+        local disable_keepalive = options.disable_keepalive or {}
+        if type(disable_keepalive) ~= 'table' then
+            error('Option disable_keepalive must be a table.')
+        end
+
         local default = {
             max_header_size     = 4096,
             header_timeout      = 100,
@@ -1297,6 +1319,7 @@ local exports = {
             log_requests        = true,
             log_errors          = true,
             display_errors      = false,
+            disable_keepalive   = {},
         }
 
         local self = {
@@ -1330,6 +1353,13 @@ local exports = {
                 ctx         = {},
                 static      = {},
             },
+
+            disable_keepalive   = tomap(disable_keepalive),
+
+            internal = {
+                preprocess_client_handler = function() end,
+                postprocess_client_handler = function() end,
+            }
         }
 
         return self
--- a/http/server.lua
+++ b/http/server.lua
@@ -758,9 +758,9 @@ local function process_client(self, s, peer)
 
         local is_eof = false
         while true do
-            local chunk = s:read{
-                delimiter = { "\n\n", "\r\n\r\n" }
-            }
+            local chunk = s:read({
+                delimiter = { "\n\n", "\r\n\r\n" },
+            }, self.idle_timeout)
 
             if chunk == '' then
                 is_eof = true
@@ -1306,6 +1306,10 @@ local exports = {
         if type(disable_keepalive) ~= 'table' then
             error('Option disable_keepalive must be a table.')
         end
+        if options.idle_timeout ~= nil and
+           type(options.idle_timeout) ~= 'number' then
+            error('Option idle_timeout must be a number.')
+        end
 
         local default = {
             max_header_size     = 4096,
@@ -1320,6 +1324,7 @@ local exports = {
             log_errors          = true,
             display_errors      = false,
             disable_keepalive   = {},
+            idle_timeout        = 0, -- no timeout, option is disabled
         }
 
         local self = {
@@ -1355,6 +1360,7 @@ local exports = {
             },
 
             disable_keepalive   = tomap(disable_keepalive),
+            idle_timeout        = options.idle_timeout,
 
             internal = {
                 preprocess_client_handler = function() end,
--- a/test/helpers.lua
+++ b/test/helpers.lua
@@ -1,6 +1,6 @@
 local fio = require('fio')
 local http_server = require('http.server')
-local http_client = require('http.client')
+local socket = require('socket')
 
 local helpers = table.copy(require('luatest').helpers)
 
@@ -87,10 +87,17 @@ helpers.find_msg_in_log_queue = function(msg, strict)
 end
 
 helpers.teardown = function(httpd)
+    local host = httpd.host
+    local port = httpd.port
     httpd:stop()
-    helpers.retrying({}, function()
-        local r = http_client.request('GET', helpers.base_uri)
-        return r == nil
+    helpers.retrying({
+        timeout = 1,
+    }, function()
+        local s, _ = socket.tcp_connect(host, port)
+        if s ~= nil then
+            s:close()
+        end
+        assert(s == nil, 'http server is stopped')
     end)
 end
 
--- a/test/integration/http_server_options_test.lua
+++ b/test/integration/http_server_options_test.lua
@@ -1,0 +1,89 @@
+local t = require('luatest')
+local http_client = require('http.client')
+
+local helpers = require('test.helpers')
+
+local g = t.group()
+
+g.after_each(function()
+    helpers.teardown(g.httpd)
+end)
+
+g.before_test('test_keepalive_is_allowed', function()
+    g.httpd = helpers.cfgserv()
+    g.httpd:start()
+end)
+
+g.test_keepalive_is_allowed = function()
+    local conn_is_opened = false
+    local conn_is_closed = false
+    g.httpd.internal.preprocess_client_handler = function() conn_is_opened = true end
+    g.httpd.internal.postprocess_client_handler = function() conn_is_closed = true end
+
+    -- Set HTTP keepalive headers: Connection:Keep-Alive and
+    -- Keep-Alive:timeout=<keepalive_idle>. Otherwise HTTP client will send
+    -- "Connection:close".
+    local opts = {
+        keepalive_idle = 3600,
+        keepalive_interval = 3600,
+    }
+    local r = http_client.new():request('GET', helpers.base_uri .. '/test', nil, opts)
+    t.assert_equals(r.status, 200)
+    t.assert_equals(r.headers.connection, 'keep-alive')
+    t.assert_equals(conn_is_opened, true)
+    t.assert_equals(conn_is_closed, false) -- Connection is alive.
+end
+
+g.before_test('test_keepalive_is_disallowed', function()
+    g.useragent = 'Mozilla/4.0'
+    g.httpd = helpers.cfgserv({
+        disable_keepalive = { g.useragent },
+    })
+    g.httpd:start()
+end)
+
+g.test_keepalive_is_disallowed = function()
+    local conn_is_opened = false
+    local conn_is_closed = false
+    g.httpd.internal.preprocess_client_handler = function() conn_is_opened = true end
+    g.httpd.internal.postprocess_client_handler = function() conn_is_closed = true end
+
+    -- Set HTTP keepalive headers: Connection:Keep-Alive and
+    -- Keep-Alive:timeout=<keepalive_idle>. Otherwise HTTP client will send
+    -- "Connection:close".
+    local opts = {
+        headers = {
+            ['user-agent'] = g.useragent,
+        },
+        keepalive_idle = 3600,
+        keepalive_interval = 3600,
+    }
+    local r = http_client.new():request('GET', helpers.base_uri .. '/test', nil, opts)
+    t.assert_equals(r.status, 200)
+    t.assert_equals(r.headers.connection, 'close')
+    t.assert_equals(conn_is_opened, true)
+    t.assert_equals(conn_is_closed, true) -- Connection is closed.
+end
+
+g.before_test('test_disable_keepalive_is_set', function()
+    g.useragent = 'Mozilla/5.0'
+    g.httpd = helpers.cfgserv({
+        disable_keepalive = { g.useragent },
+    })
+    g.httpd:start()
+end)
+
+g.test_disable_keepalive_is_set = function(g)
+    local httpd = g.httpd
+    t.assert_equals(httpd.disable_keepalive[g.useragent], true)
+end
+
+g.before_test('test_disable_keepalive_default', function()
+    g.httpd = helpers.cfgserv()
+    g.httpd:start()
+end)
+
+g.test_disable_keepalive_default = function(g)
+    local httpd = g.httpd
+    t.assert_equals(table.getn(httpd.options.disable_keepalive), 0)
+end
--- a/test/integration/http_server_options_test.lua
+++ b/test/integration/http_server_options_test.lua
@@ -87,3 +87,81 @@ g.test_disable_keepalive_default = function(g)
     local httpd = g.httpd
     t.assert_equals(table.getn(httpd.options.disable_keepalive), 0)
 end
+
+g.before_test('test_idle_timeout_default', function()
+    g.httpd = helpers.cfgserv()
+    g.httpd:start()
+end)
+
+g.test_idle_timeout_default = function(g)
+    local httpd = g.httpd
+    t.assert_equals(httpd.options.idle_timeout, 0)
+end
+
+g.before_test('test_idle_timeout_is_set', function()
+    g.idle_timeout = 0.5
+    g.httpd = helpers.cfgserv({
+        idle_timeout = g.idle_timeout,
+    })
+    g.httpd:start()
+end)
+
+g.test_idle_timeout_is_set = function(g)
+    local conn_is_opened = false
+    local conn_is_closed = false
+    local httpd = g.httpd
+    g.httpd.internal.preprocess_client_handler = function() conn_is_opened = true end
+    g.httpd.internal.postprocess_client_handler = function() conn_is_closed = true end
+
+    t.assert_equals(httpd.options.idle_timeout, g.idle_timeout)
+
+    -- Set HTTP keepalive headers: Connection:Keep-Alive and
+    -- Keep-Alive:timeout=<keepalive_idle>. Otherwise HTTP client will send
+    -- "Connection:close".
+    local opts = {
+        keepalive_idle = 3600,
+        keepalive_interval = 3600,
+    }
+    local r = http_client.new():request('GET', helpers.base_uri .. '/test', nil, opts)
+    t.assert_equals(r.status, 200)
+    t.assert_equals(r.headers.connection, 'keep-alive')
+    t.assert_equals(conn_is_opened, true)
+
+    helpers.retrying({
+        timeout = g.idle_timeout * 2,
+    }, function()
+        assert(conn_is_closed, 'connection is closed')
+    end)
+    t.assert_equals(conn_is_closed, true) -- Connection is closed.
+end
+
+g.before_test('test_idle_timeout_is_unset', function()
+    g.httpd = helpers.cfgserv({})
+    g.httpd:start()
+end)
+
+g.test_idle_timeout_is_unset = function(g)
+    local conn_is_opened = false
+    local conn_is_closed = false
+    g.httpd.internal.preprocess_client_handler = function() conn_is_opened = true end
+    g.httpd.internal.postprocess_client_handler = function() conn_is_closed = true end
+
+    -- Set HTTP keepalive headers: Connection:Keep-Alive and
+    -- Keep-Alive:timeout=<keepalive_idle>. Otherwise HTTP client will send
+    -- "Connection:close".
+    local opts = {
+        keepalive_idle = 3600,
+        keepalive_interval = 3600,
+    }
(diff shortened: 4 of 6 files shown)


README.md
```diff
@@ -130,6 +130,9 @@ httpd = require('http.server').new(host, port[, { options } ])
 
   By default uses `log.info` function for requests logging.
 * `log_errors` - same as the `log_requests` option but is used for error messages logging. By default uses `log.error()` function.
+* `keepalive_disable` - disables keep-alive connections with misbehaving
```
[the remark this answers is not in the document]
ligurio: > What's the motivation behind this solution?

Simplicity. There was a problem on production with exactly a single HTTP client.
My pull request solves the problem described in a ticket.
Regarding choosing map or an array: the solution with map is much easy to implement. With array, as you correctly said you need to convert to a map. Then you need to put a map to cache and correctly invalidate cache because user can change `httpd.options.keepalive_disable` in runtime.

> Moreover, this solution should be thoughtfully verified with customer @vasiliy-t . I'm not sure it will satisfy Delivery team.

Sure, it should be. When we will merge this change to a master. Vasya needs a fix as a Luarocks release for fix validation.

README.md
```diff
@@ -130,6 +130,9 @@ httpd = require('http.server').new(host, port[, { options } ])
...
+  clients. Parameter accept a map that contains a user agents with non-nil
```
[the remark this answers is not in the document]
ligurio: The solution with map is much easier to implement. With array, as you correctly said you need to convert to a map. Then you need to put a map to cache and correctly invalidate cache because user can change `httpd.options.keepalive_disable` in runtime.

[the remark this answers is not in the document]
ligurio: Added example to the README.
```patch
@@ -130,6 +130,22 @@ httpd = require('http.server').new(host, port[, { options } ])
 
   By default uses `log.info` function for requests logging.
 * `log_errors` - same as the `log_requests` option but is used for error messages logging. By default uses `log.error()` function.
+* `keepalive_disable` - disables keep-alive connections with misbehaving
+  clients. Parameter accept a map that contains a user agents with non-nil
+  value. By default map is empty.
+
+  Example:
+
+  ```lua
+  local httpd = http_server.new('127.0.0.1', 8080, {
+      log_requests = true,
+      log_errors = true,
+      keepalive_disable = {
+          ['curl/7.68.0'] = true
+      },
+  })
+  ```
+
 
 ## Using routes
 
```

README.md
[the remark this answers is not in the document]
DifferentialOrange: Well, when this will be merged to master and tagged, it would be too late to change implementation and verify anything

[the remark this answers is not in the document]
DifferentialOrange: Implementation of solution is LGTM. I can't say that I do not have doubts about solution design itself, so I propose to ask someone to review it additionally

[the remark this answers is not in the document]
DifferentialOrange: AFAIK, the result of the discussion with Delivery was to keep this feature for now and to make something more forward-thinking in future. I don't see any blocks for this PR. I leave the decision about closing #137 ticket with this PR and opening a new one for another feature (like killing idle keep-alive connections) or keeping #137 opened to author.

DifferentialOrange (approved):

http/server.lua
```diff
@@ -23,6 +23,17 @@ local function sprintf(fmt, ...)
     return string.format(fmt, ...)
 end
 
+local function is_no_keepalive(disallowed_map, useragent)
+    if type(disallowed_map) ~= 'table' or useragent == nil then
```
olegrok, line 27: I guess we could check that "disallowed_map" on initialization step. And here two variants are possible - nil or table.
It seems to be hot path and I believe nil check is faster than type check.

http/server.lua
```diff
@@ -23,6 +23,17 @@ local function sprintf(fmt, ...)
...
+        return false
+    end
+    if disallowed_map[useragent] ~= nil then
```
olegrok, line 30: I think here should be "true" check.
I see on example:
```
keepalive_disable = {
          ['curl/7.68.0'] = true
      },
```

But what if I write:
```
keepalive_disable = {
          ['curl/7.68.0'] = false
      },
```
?

vasiliy-t:
I does not agree with this pr. Any client could misbehave and leave open connections, every client could change it's user agent. I suggest implementing idle timeout option to set hard limit on connections time to live as more generic and robust solution.

http/server.lua
ligurio: Fixed.

```lua
--- a/http/server.lua
+++ b/http/server.lua
@@ -24,10 +24,10 @@ local function sprintf(fmt, ...)
 end
 
 local function is_no_keepalive(disallowed_map, useragent)
-    if type(disallowed_map) ~= 'table' or useragent == nil then
+    if useragent == nil then
         return false
     end
-    if disallowed_map[useragent] ~= nil then
+    if disallowed_map[useragent] == true then
         return true
     end
```

http/server.lua
ligurio:
```lua
@@ -1299,6 +1299,11 @@ local exports = {
         if type(options) ~= 'table' then
             errorf("options must be table not '%s'", type(options))
         end
+        options.keepalive_disable = options.keepalive_disable or {}
+        if type(options.keepalive_disable) ~= 'table' then
+            error('Option keepalive_disable must be a table.')
+        end
+
         local default = {
             max_header_size     = 4096,
             header_timeout      = 100,

```

ligurio:
Added draft of patch with option `idle_timeout`. @olegrok, could you review it?

http/server.lua
```diff
@@ -15,6 +16,21 @@ local errno = require 'errno'
...
 local DETACHED = 101
 
+-- A table that maps fiber object to time when it was active last time.
+-- Using for tracking idle connections.
+local connections = {}
+
+local function close_idle_connections(connections, idle_timeout)
+    -- FIXME: complexity is O(N) and it is bad
+    for fiber_obj, start_time in pairs(connections) do
+        if fiber.clock() - start_time >= idle_timeout then
```
olegrok, line 26: Probably we could calculate this value only once at start of the loop.
And also you can use queue instead of array here to solve fixme.

http/server.lua
```diff
@@ -15,6 +16,21 @@ local errno = require 'errno'
...
+            log.info("Connection closed, idle_timeout was exceeded (fiber id %d)\n", fiber_obj:id())
```
olegrok, line 27: We have log_errors and log_requests callbacks. So I guess it's a bit inconsistent to use pure built-in log.

http/server.lua
```diff
@@ -962,6 +1000,9 @@ local function httpd_stop(self)
         self.tcp_server:close()
         self.tcp_server = nil
     end
+
+    connections = {}
```
olegrok, line 1004: It looks it should be a property of sever object. As I understand currently stop of one server will affect other servers (user could create several servers).

http/server.lua
```diff
@@ -746,6 +773,11 @@ local function process_client(self, s, peer)
     while true do
         local hdrs = ''
 
+        connections[fiber.id] = fiber.clock()
```
olegrok, line 776: I guess if options is not enabled we shouldn't call fiber.clock at all.

ligurio: Fixed.

```lua
--- a/http/server.lua
+++ b/http/server.lua
@@ -773,8 +773,8 @@ local function process_client(self, s, peer)
     while true do
         local hdrs = ''
 
-        connections[fiber.id] = fiber.clock()
         if self.options.idle_timeout > 0 then
+            connections[fiber.id] = fiber.clock()
             close_idle_connections(connections, self.options.idle_timeout)
         end
 
```

http/server.lua
ligurio: Done.

```lua
diff --git a/http/server.lua b/http/server.lua
index 84f0f24..6b58f9d 100644
--- a/http/server.lua
+++ b/http/server.lua
@@ -16,10 +16,6 @@ local errno = require 'errno'
 
 local DETACHED = 101
 
--- A table that maps fiber object to time when it was active last time.
--- Using for tracking idle connections.
-local connections = {}
-
 local function close_idle_connections(connections, idle_timeout)
     -- FIXME: complexity is O(N) and it is bad
     for fiber_obj, start_time in pairs(connections) do
@@ -775,7 +771,7 @@ local function process_client(self, s, peer)
 
         if self.options.idle_timeout > 0 then
             connections[fiber.id] = fiber.clock()
-            close_idle_connections(connections, self.options.idle_timeout)
+            close_idle_connections(self.connections, self.options.idle_timeout)
         end
 
         local is_eof = false
@@ -1001,7 +997,7 @@ local function httpd_stop(self)
         self.tcp_server = nil
     end
 
-    connections = {}
+    self.connections = {}
 
     return self
 end
@@ -1387,6 +1383,10 @@ local exports = {
                 ctx         = {},
                 static      = {},
             },
+
+            -- A table that maps fiber object to time when it was active last
+            -- time. Using for tracking idle connections.
+            connections = {},
         }
 
         return self
```

http/server.lua
ligurio: Done.

```lua
diff --git a/http/server.lua b/http/server.lua
index 6b58f9d..5b80de5 100644
--- a/http/server.lua
+++ b/http/server.lua
@@ -20,7 +20,7 @@ local function close_idle_connections(connections, idle_timeout)
     -- FIXME: complexity is O(N) and it is bad
     for fiber_obj, start_time in pairs(connections) do
         if fiber.clock() - start_time >= idle_timeout then
-            log.info("Connection closed, idle_timeout was exceeded (fiber id %d)\n", fiber_obj:id())
+            logreq("Connection closed, idle_timeout was exceeded (fiber id %d)\n", fiber_obj:id())
             fiber_obj:cancel()
             connections[fiber_obj] = nil
         end
@@ -768,10 +768,12 @@ end
 local function process_client(self, s, peer)
     while true do
         local hdrs = ''
+        local route = self:match(p.method, p.path)
+        local logreq = get_request_logger(self.options, route)
 
         if self.options.idle_timeout > 0 then
             connections[fiber.id] = fiber.clock()
-            close_idle_connections(self.connections, self.options.idle_timeout)
+            close_idle_connections(self.connections, self.options.idle_timeout, logreq)
         end
 
         local is_eof = false
@@ -815,8 +817,6 @@ local function process_client(self, s, peer)
             s:write('HTTP/1.0 100 Continue\r\n\r\n')
         end
 
-        local route = self:match(p.method, p.path)
-        local logreq = get_request_logger(self.options, route)
         logreq("%s %s%s", p.method, p.path,
                p.query ~= "" and "?"..p.query or "")
 ```

File: http/server.lua (hunk 8 above)
Comment (olegrok) [inline http/server.lua:776]: "fiber.id" is a function. Not just numeric value. Probably it should be `fiber.id()`, anyway how does it pass tests?

Comment (ligurio) [reply http/server.lua]:
```lua
--- a/http/server.lua
+++ b/http/server.lua
@@ -774,7 +774,8 @@ local function process_client(self, s, peer)
         local logger = get_request_logger(self.options)
 
         if self.options.idle_timeout > 0 then
-            connections[fiber.id] = fiber.clock()
+            local id = fiber.id()
+            connections[id] = fiber.clock()
             close_idle_connections(self.connections, self.options.idle_timeout, logger)
         end
 
```

>  anyway how does it pass tests?

I don't know a reliable way to test this feature and tests are simple right now.

ligurio: In fact, I need a fiber object, not a fiber id. Because `close_idle_connections()` should somehow call `fiber:cancel()`.

http/server.lua
```diff
@@ -742,7 +754,43 @@ local function parse_request(req)
...
+    }
+
+    return true
+end
+
+local function close_connection(self, fiber_obj, fiber_id, logger)
+    -- Name of fiber for http connection contains a hostname and port of http client.
+    local fiber_name = fiber_obj:name()
+    self.connections[fiber_id] = nil
+    logger("Close connection with %s, idle_timeout was exceeded\n", fiber_name)
```
olegrok, line 771: Not sure that fiber_name will say anything. Probably we could put in message `"%s (%d) - %s", fiber.name(), fiber.id(), peer`

test/integration/http_server_options_test.lua
```diff
@@ -0,0 +1,75 @@
...
+
+g.test_keepalive_change_in_runtime = function(g)
+    local httpd = g.httpd
+    t.assert_equals(httpd.options.keepalive_disable, {})
+    httpd.options.keepalive_disable = {
+        ['Mozilla/4.0'] = true
+    }
+    t.assert_equals(httpd.options.keepalive_disable, {
+        ["Mozilla/4.0"] = true
+    })
```
Totktonada, line 88: It is a bit strange. We assign into a table field and assert that it is there. Shouldn't we test that it is in effect?

README.md
```diff
@@ -132,6 +132,22 @@ httpd = require('http.server').new(host, port[, { options } ])
...
+
+  Example:
+
+  ```lua
+  local httpd = http_server.new('127.0.0.1', 8080, {
+      log_requests = true,
+      log_errors = true,
+      keepalive_disable = {
+          ['curl/7.68.0'] = true
+      },
```
Totktonada, line 145: I don't know, whether it was already discussed or not, but `{['foo'] = true, ['bar'] = true}` is logical equivalent of `{'foo', 'bar'}`. I would expose the latter variant externally and transform it to the former internally.

I don't insist, but it looks a bit more accurate for me.

http/server.lua
```diff
@@ -753,7 +754,43 @@ local function parse_request(req)
...
+            self:close_connection(opts.obj, fiber_id, logger)
+        end
+    end
+end
+
 local function process_client(self, s, peer)
+    if self.options.idle_timeout > 0 then
+        self:check_connections()
+        self:add_connection(fiber.self())
+    end
```
Totktonada, line 792: This way we'll cancel long requests (when a client's code doing a long work) and slow requests (say, when writting a large response requires considerable time). Not only idle connections.

It seems, we can just set the `idle_timeout` as timeout for the `:read()` call several lines below and handle the timeout error properly.

http/server.lua
```diff
@@ -1344,6 +1389,10 @@ local exports = {
             hook    = set_hook,
             url_for = url_for_httpd,
 
+            add_connection = add_connection,
+            close_connection = close_connection,
+            check_connections = check_connections,
```
Totktonada, line 1394: Redefinition of those functions looks dangerous. I would not expose them.

http/server.lua
ligurio: Fiber name that processes connection in `socket.tcp_server` ([documentation](https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/#socket-tcp-connect)) is not a plain string, but a host IP address and port. See snippet below:

```lua
sergeyb@pony:~/sources/MRG/http$ cat fiber_name.lua 
local socket = require('socket')
local fiber = require('fiber')

local function handler()
    print('Fiber name is', fiber.self():name())
end
socket.tcp_server('localhost', 8000, handler)
socket.tcp_connect('localhost', 8000)
sergeyb@pony:~/sources/MRG/http$ tarantool fiber_name.lua 
started
Fiber name is   server/127.0.0.1:44634
entering the event loop
```

README.md
ligurio: I like idea. Patch updated.

http/server.lua
ligurio: Redefinition? What do you mean?
It is a definition of internal methods required for testing of idle connections.

Totktonada: At least start them from underscore, it'll exclude the fields from autocompletion.

(However if you'll follow my proposal regarding `:read()` timeout -- and I'm sure it is worthful -- those functions will gone.)

http/server.lua
ligurio: Patch updated (not squashed yet).

http/server.lua
ligurio: Non-relevant functions removed (`add_connection`, `close_connection`, `check_connections`).

test/integration/http_server_options_test.lua
```diff
@@ -0,0 +1,170 @@
...
+    t.assert_equals(response.status, 200)
+    t.assert_equals(response.headers.connection, 'keep-alive')
+    t.assert_equals(httpd.num_connections, 1)
+
+    -- Wait for IDLE_TIMEOUT, after that connection must be closed by timeout.
+    require('fiber').sleep(IDLE_TIMEOUT)
+
+    -- Server will handle http request using previously started fiber if
+    -- connection was not closed. If connection was closed then server will
+    -- start new fiber with handler that it will increment num_connections.
```
opomuc, line 131: how can we be sure, that a new connection is created, if we always check only for `num_connections == 1`?

ligurio: Using internal handlers.

http/server.lua
ligurio: Patch was rewritten (I'm using `timeout` in `tcp_server()` as you suggested before).

http/server.lua
ligurio: > This way we'll cancel long requests (when a client's code doing a long work) and slow requests (say, when writting a large response requires considerable time). Not only idle connections.

Right, implementation was not ideal. It was an attempt to create a scheduler for cooperative multitasking.

Patch was rewritten and now uses `timeout` option in `tcp_server`.

test/integration/http_server_options_test.lua
ligurio: Now it is tested with internal handlers.

http/server.lua
ligurio: Rewrote patch with using `timeout` in `tcp_server`.

http/server.lua
Totktonada: Now you added the `internal` field instead of the functions. It should be started with underscore to don't participate in tab-completion.

http/server.lua
ligurio: > Now you set the timeout for getaddinfo() at socket binding. It is completely unrelated here.

Right. https://github.com/tarantool/doc/issues/2946

test/helpers.lua
```diff
@@ -94,4 +96,47 @@ helpers.teardown = function(httpd)
...
+function helpers.wait_cond(cond, timeout, delay)
+    assert(type(cond) == 'function')
+
+    local timeout = timeout or 60
+    local delay = delay or 0.001
+
+    local start_time = clock.monotonic()
+    local res = {
+		pcall(cond),
+	}
```
Totktonada, line 126: Tab/space mix.

test/helpers.lua
```diff
@@ -94,4 +96,47 @@ helpers.teardown = function(httpd)
...
+
+    while not res[1] or not res[2] do
+        local work_time = clock.monotonic() - start_time
+        if work_time > timeout then
+            return wait_cond_finish(res[1], res[2])
+        end
+        fiber.sleep(delay)
+        res = {
+			pcall(cond),
+		}
```
Totktonada, line 136: Tab/space mix.

It is easy to see with refined-github web browser plugin.

http/server.lua
Totktonada: > The 'test_idle_timeout' test passes, because it does not test anything. The keepalive_* options are not set for the http client.

It is still so.

```diff
diff --git a/http/server.lua b/http/server.lua
index 08d59b5..3f8c887 100644
--- a/http/server.lua
+++ b/http/server.lua
@@ -762,7 +762,7 @@ local function process_client(self, s, peer)
                 delimiter = { "\n\n", "\r\n\r\n" },
                 -- Maximum number of seconds to wait,
                 -- for example 50 means "stop after 50 seconds".
-                timeout = self.idle_timeout,
+                -- timeout = self.idle_timeout,
             }
 
             if chunk == '' then
```

```shell
$ cmake . && make && luatest -v
<...>
Ran 54 tests in 0.543 seconds, 54 succeeded, 0 failed
```

0x501D (approved):
LGTM

http/server.lua
Totktonada: Okay.

test/helpers.lua
```diff
@@ -87,10 +87,14 @@ helpers.find_msg_in_log_queue = function(msg, strict)
...
     httpd:stop()
-    helpers.retrying({}, function()
-        local r = http_client.request('GET', helpers.base_uri)
-        return r == nil
+    helpers.retrying({
+        timeout = 1,
+    }, function()
+        local s, _ = socket.tcp_connect(host, port)
+        return s == nil
     end)
```
Totktonada, line 98: `helpers.retrying()` retries when its function *raises* an error. For example:

```diff
diff --git a/test/helpers.lua b/test/helpers.lua
index a9a80aa..9f33068 100644
--- a/test/helpers.lua
+++ b/test/helpers.lua
@@ -94,7 +94,7 @@ helpers.teardown = function(httpd)
         timeout = 1,
     }, function()
         local s, _ = socket.tcp_connect(host, port)
-        return s == nil
+        assert(s == nil, 'assert that the http server is stopped')
     end)
 end
```

It would be even better to don't lean on GC in a frequest operation that opens a new file descriptor (to don't hit the limit):

```diff
diff --git a/test/helpers.lua b/test/helpers.lua
index a9a80aa..0ee0bf6 100644
--- a/test/helpers.lua
+++ b/test/helpers.lua
@@ -94,7 +94,10 @@ helpers.teardown = function(httpd)
         timeout = 1,
     }, function()
         local s, _ = socket.tcp_connect(host, port)
-        return s == nil
+        if s ~= nil then
+            s:close()
+        end
+        assert(s == nil, 'assert that the http server is stopped')
     end)
 end
```

test/integration/http_server_options_test.lua
```diff
@@ -87,3 +87,80 @@ g.test_disable_keepalive_default = function(g)
...
+    t.assert_equals(r.headers.connection, 'keep-alive')
+    t.assert_equals(conn_is_opened, true)
+
+    helpers.retrying({
+        timeout = g.idle_timeout * 2,
+    }, function()
+        if conn_is_closed ~= true then
+            assert()
+        end
+    end)
```
Totktonada, line 136: It can be simpler:

```diff
diff --git a/test/integration/http_server_options_test.lua b/test/integration/http_server_options_test.lua
index 504891d..e5411a6 100644
--- a/test/integration/http_server_options_test.lua
+++ b/test/integration/http_server_options_test.lua
@@ -130,9 +130,7 @@ g.test_idle_timeout_is_set = function(g)
     helpers.retrying({
         timeout = g.idle_timeout * 2,
     }, function()
-        if conn_is_closed ~= true then
-            assert()
-        end
+        assert(conn_is_closed, 'assert that the connection is closed')
     end)
     t.assert_equals(conn_is_closed, true) -- Connection is closed.
 end
```

test/integration/http_server_options_test.lua
```diff
@@ -87,3 +87,80 @@ g.test_disable_keepalive_default = function(g)
...
+        keepalive_idle = 3600,
+        keepalive_interval = 3600,
+    }
+    local r = http_client.new():request('GET', helpers.base_uri .. '/test', nil, opts)
+    t.assert_equals(r.status, 200)
+    t.assert_equals(r.headers.connection, 'keep-alive')
+    t.assert_equals(conn_is_opened, true)
+    helpers.retrying({}, function()
+        return conn_is_closed == false
+    end)
```
Totktonada, line 164: The same as for `helpers.retrying()` in `helpers.teardown()`: we should *raise* an error when we don't meet an expected condition.

```diff
@@ -160,7 +158,7 @@ g.test_idle_timeout_is_unset = function(g)
     t.assert_equals(r.headers.connection, 'keep-alive')
     t.assert_equals(conn_is_opened, true)
     helpers.retrying({}, function()
-        return conn_is_closed == false
+        assert(conn_is_closed == false, 'assert that the connection is closed')
     end)
     t.assert_equals(conn_is_closed, false) -- Connection is alive.
 end
```

Totktonada: Looked a bit deeper into the test case. `conn_is_closed` is `false` initially, there is no sense to wait for this situation: `helpers.retrying()` will always return immediately. We can either check that the connection was not closed by the server immediately or wait for some short time and check.

For example:

```diff
@@ -159,8 +157,5 @@ g.test_idle_timeout_is_unset = function(g)
     t.assert_equals(r.status, 200)
     t.assert_equals(r.headers.connection, 'keep-alive')
     t.assert_equals(conn_is_opened, true)
-    helpers.retrying({}, function()
-        return conn_is_closed == false
-    end)
     t.assert_equals(conn_is_closed, false) -- Connection is alive.
 end
```

Or:

```diff
diff --git a/test/integration/http_server_options_test.lua b/test/integration/http_server_options_test.lua
index 504891d..ef2954f 100644
--- a/test/integration/http_server_options_test.lua
+++ b/test/integration/http_server_options_test.lua
@@ -1,3 +1,4 @@
+local fiber = require('fiber')
 local t = require('luatest')
 local http_client = require('http.client')
 

@@ -158,9 +157,12 @@ g.test_idle_timeout_is_unset = function(g)
     local r = http_client.new():request('GET', helpers.base_uri .. '/test', nil, opts)
     t.assert_equals(r.status, 200)
     t.assert_equals(r.headers.connection, 'keep-alive')
+
+    -- The server should not close the connection during the
+    -- keepalive_idle seconds. We can't wait so long in the test,
+    -- so let's wait for a short time and check.
+    fiber.sleep(0.5)
+
     t.assert_equals(conn_is_opened, true)
-    helpers.retrying({}, function()
-        return conn_is_closed == false
-    end)
     t.assert_equals(conn_is_closed, false) -- Connection is alive.
 end
```

Totktonada (approved):
LGTM aside of several nits regarding the tests.

Totktonada:
Pavel blessed the push in a private conversation.

PR #146, opened 2021-10-29
Disable option display_errors by default

### LINKED ISSUE #129: Readme is wrong
Given examples don't work.

For example: https://github.com/tarantool/http/blob/master/README.md#working-with-stashes

```lua
httpd = box.httpd.new('127.0.0.1', 8080)
...
```

There is no such thing as box.httpd.

### LINKED ISSUE #135: URL in README.md is broken
https://raw.githubusercontent.com/tarantool/http/master/rockspecs/http-scm-1.rockspec
404

ligurio:
Fixes https://github.com/tarantool/security/issues/8
Fixes #129 
Fixes #135 
Fixes a CMake warnings for passed package names

~~**PR depends on https://github.com/tarantool/http/pull/141**~~

NOTE: Issue https://github.com/tarantool/security/issues/8 must be closed manually after merging.

### DIFF
--- a/http/server.lua
+++ b/http/server.lua
@@ -1289,7 +1289,7 @@ local exports = {
             cache_static        = true,
             log_requests        = true,
             log_errors          = true,
-            display_errors      = true,
+            display_errors      = false,
         }
 
         local self = {
--- a/http/server.lua
+++ b/http/server.lua
@@ -859,6 +859,10 @@ local function process_client(self, s, peer)
             hdrs['content-length'] = #body
         end
 
+        if hdrs['content-type'] == nil then
+            hdrs['content-type'] = 'text/plain; charset=utf-8'
+        end
+
         if hdrs.server == nil then
             hdrs.server = sprintf('Tarantool http (tarantool v%s)', _TARANTOOL)
         end
--- a/http/server.lua
+++ b/http/server.lua
@@ -1328,6 +1328,7 @@ local exports = {
     internal = {
         response_mt = response_mt,
         request_mt = request_mt,
+        extend = extend,
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- Disable option display_errors by default.
+
 ## [1.1.1] - 2021-10-28
 
 ### Changed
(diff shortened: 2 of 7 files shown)


DifferentialOrange (approved):
I think it is worth to duplicate commit messages (like the one about Django) additional info to PR description since public users do not have access to tarantool/security ticket.

ligurio:
> I think it is worth to duplicate commit messages (like the one about Django) additional info to PR description since public users do not have access to tarantool/security ticket.

It was already there, except link to site about django. Well, added it too.

ligurio:
Screenshot from 2021-11-02 15-47-39

DifferentialOrange:
> > I think it is worth to duplicate commit messages (like the one about Django) additional info to PR description since public users do not have access to tarantool/security ticket.
> 
> It was already there, except link to site about django. Well, added it too.

I mean this description
image

ligurio:
>>> I think it is worth to duplicate commit messages (like the one about Django) additional info to PR description since public users do not have access to tarantool/security ticket.

>> It was already there, except link to site about django. Well, added it too.

> I mean this description

What's a point to publish a description of security issue before fixing it?

ligurio:
@DifferentialOrange I added to PR description all issues that fixed by this PR. Also removed commits related to #137, so we can merge these commits without waiting of fix for #137.

http/server.lua
```diff
@@ -859,6 +859,10 @@ local function process_client(self, s, peer)
             hdrs['content-length'] = #body
         end
 
+        if hdrs['content-type'] == nil then
+            hdrs['content-type'] = 'text/plain; charset=utf-8'
```
Totktonada, line 863: Nit: It is 'text/plain; utf-8' in the commit message, looks as the typo.

http/server.lua
```diff
@@ -1290,6 +1290,7 @@ local exports = {
             log_requests        = true,
             log_errors          = true,
             display_errors      = true,
+            keepalive_disable   = {},
```
Totktonada, line 1297: It does not look related to this commit.

http/server.lua
Totktonada: I would also leave several words why default content type may be important from security point of view.

Totktonada (approved):
Looks good except several minor nits. Feel free to push after fixes.

http/server.lua
ligurio: Fixed.

http/server.lua
ligurio: True, fixed.

ligurio:
> I would also leave several words why default content type may be important from security point of view.

Updated commit message.

PR #155, opened 2021-11-10
Remove rockspecs

ligurio:
We publish rockspecs and rocks on Tarantool's rock server [1]. No sense
to duplicate them in project's VCS. Rockspec scm-1 will be always here
and it's easy to make a rockspec for a certain version using it.

1. https://rocks.tarantool.org/


Totktonada (approved):
Together with such change we usually move the scm-1 rockspec to the repository root: https://github.com/tarantool/avro-schema/pull/80

Aside of this, I have no objections. Feel free to push.

PR #158, opened 2022-01-22
Update server.lua

a1div0:
The `tcp_server` socket method call has been separated into a separate http `listen` method so that `listen` can be overridden and use a different socket. Namely: [sslsocket](https://github.com/tarantool/sslsocket). To make a `https` server.
The call to require('socket') has been moved for the same reason:
1. The call itself is rare.
2. It is require('socket') that may not be needed if `listen` is overridden.
---
Вызов метода сокета `tcp_server` выделен в отдельный метод http `listen` для того, чтобы `listen` можно было переопределить и использовать другой сокет. А именно: [sslsocket](https://github.com/tarantool/sslsocket). Чтобы сделать `https` сервер.
Вызов require('socket') перемещён по той же причине:
1. Сам вызов происходит редко.
2. Именно require('socket') может не понадобиться, если `listen` будет переопределён.

### DIFF
--- a/http/server.lua
+++ b/http/server.lua
@@ -9,7 +9,6 @@ local mime_types = require('http.mime_types')
 local codes = require('http.codes')
 
 local log = require('log')
-local socket = require('socket')
 local json = require('json')
 local errno = require 'errno'
 
@@ -1251,16 +1250,26 @@ local function url_for_httpd(httpd, name, args, query)
     end
 end
 
+local function tcp_server_create(self)
+    local socket = require('socket')
+    return socket.tcp_server(
+        self.host
+        ,self.port
+        ,{
+            name = 'http',
+            handler = function(...)
+                process_client(self, ...)
+            end
+        }
+    )
+end
+
 local function httpd_start(self)
     if type(self) ~= 'table' then
         error("httpd: usage: httpd:start()")
     end
 
-    local server = socket.tcp_server(self.host, self.port,
-                     { name = 'http',
-                       handler = function(...)
-                           process_client(self, ...)
-                     end})
+    local server = self:listen()
     if server == nil then
         error(sprintf("Can't create tcp_server: %s", errno.strerror()))
     end
@@ -1317,6 +1326,7 @@ local exports = {
             helper  = set_helper,
             hook    = set_hook,
             url_for = url_for_httpd,
+            listen  = tcp_server_create,
 
             -- caches
             cache   = {


a1div0:
Данное изменение никак не меняет описание использования модуля, так как функциональность модуля не меняется. Но возможности по использованию увеличиваются. Пожалуйста посмотрите мои изменения.
---
This change does not change the description of the use of the module in any way, since the functionality of the module does not change. But the possibilities for use are increasing. Please see my changes.

Totktonada:
Hi The idea is clear enough and looks meaningful.

Several general comments:

1. Just English is enough. No need to duplicate description on other languages.
2. What problem you aims to solve by moving `require('socket')`? Even if we don't use it, it is harmless to require it as usual: at top of the file.
3. Please, follow our [Lua style guide](https://www.tarantool.io/en/doc/latest/dev_guide/lua_style_guide/) (at least, don't use line-start-comma `, item` style and prefer line-end-comma `item,`) and [commit message recommendations](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message) (at least the title should be a conscise description of your change rather than what file is changed; also avoid long lines in the text).
4. It would be nice to add a simple regression test, add the new method into README and leave a note in CHANGELOG.

I'll look more close after this.

a1div0:
1. Ok
2. Moved require('socket') inside the procedure, because it refers to it. If the procedure is overridden, then the call is not needed. Just redundant.
On the other hand, placing on top makes sense for understanding which modules are needed, or potentially needed. I don't know how better. Ready to redo it, of course.
3. Understood. I'm studying. I wonder if you don't have "IntelliJ IDEA code style XML" (settings -> Editor -> Code Style Scheme -> Export...)? I would just upload it to myself, yes and all.
4. Ok I will do

Totktonada:
A suggestion for future changes: you could force-push (`git push -f`) your branch and update PR title/description instead of closing an old PR and opening a new one. This way the discussion will not spread across several web pages and is easier to read.

2. I would suggest to keep it unchanged (on top of the module). The rule of thumb is to don't do unnecessary changes. The practical sense here is the following: I know that requires usually at top of a file and exceptions are usually motivated either by avoiding cyclic dependencies or (rarely) by compatibility matters (say, Lua 5.1 / Lua 5.3 / LuaJIT 2.1 / Tarantool). So a reader of the changed code may think that there is a trick here, but it is not so.
3. No, I have no code style file for IDEA.

PR #159, opened 2022-01-22
Added the ability to use a non-standard socket (for example, with support   for SSL)

a1div0:
In continuation [pull 158](https://github.com/tarantool/http/pull/158).
Please review and accept the changes. Admonitions have been taken into account.
Summary:
1. The `tcp_server` socket method call has been separated into a separate http `listen` method so that `listen` can be overridden and use a different socket.
2. `requirement('socket')` moved inside procedure as it is only referenced by the default implementation. If the procedure is overridden, then the call is not needed. Just too much.
On the other hand, placement on top makes sense for understanding what modules are needed or potentially needed. I don't know which is better. Ready to redo, of course.
3. Fixed syntax
4. Add info to README
5. Add unit test (for regress testing)
6. Add info to CHANGELOG

### DIFF
--- a/http/server.lua
+++ b/http/server.lua
@@ -1256,11 +1256,14 @@ local function httpd_start(self)
         error("httpd: usage: httpd:start()")
     end
 
-    local server = socket.tcp_server(self.host, self.port,
-                     { name = 'http',
-                       handler = function(...)
-                           process_client(self, ...)
-                     end})
+    local server = self.tcp_server_f(self.host, self.port, {
+        name = 'http',
+        handler = function(...)
+            process_client(self, ...)
+        end,
+        http_server = self,
+    })
+
     if server == nil then
         error(sprintf("Can't create tcp_server: %s", errno.strerror()))
     end
@@ -1317,6 +1320,9 @@ local exports = {
             helper  = set_helper,
             hook    = set_hook,
             url_for = url_for_httpd,
+
+            -- Exposed to make it replaceable by a user.
+            tcp_server_f = socket.tcp_server,
 
             -- caches
             cache   = {
--- a/test/unit/http_custom_socket_test.lua
+++ b/test/unit/http_custom_socket_test.lua
@@ -1,0 +1,50 @@
+local t = require('luatest')
+local http_server = require('http.server')
+
+local g = t.group()
+
+g.test_custom_socket = function()
+    local is_listening = false
+
+    local function tcp_server(host, port, opts)
+        assert(type(opts) == 'table')
+        local name = opts.name
+        local accept_handler = opts.handler
+        local http_server = opts.http_server
+
+        -- Verify arguments.
+        t.assert_equals(host, 'host', 'check host')
+        t.assert_equals(port, 123, 'check port')
+
+        -- Verify options.
+        t.assert_equals(name, 'http', 'check server name')
+        t.assert_type(accept_handler, 'function', 'check accept handler')
+        t.assert_type(http_server.routes, 'table',
+            'http server object is accessible')
+
+        is_listening = true
+        return {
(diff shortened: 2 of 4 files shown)


Totktonada:
Made several changes (force-pushed):

1. Squashed commits, clarified commit title/message.
2. Reformat readme text to 80 columns, fix typos.
3. Code style, naming, luacheck fixes in the test.

Totktonada:
I also added one more commit, which proposes another way to redefine the socket implementation. It looks more straightforward and don't reveals implementation details of the http server module (a user shouldn't discard the first argument — http server object — at least). It is to be discussed and either discarded or squashed.

@ligurio Please, take a look, when you'll have a spare time.

a1div0:
Thanks! So it got better.

test/unit/http_custom_socket_test.lua
```diff
@@ -0,0 +1,50 @@
...
+    -- that the HTTP server sets `myhttp.tcp_server` to the TCP
+    -- server object. The name clash is harmless for the module,
+    -- but may confuse a user, so the `tcp_server_f` was chosen
+    -- instead of just `tcp_server`. The `tcp_server_f` field is
+    -- guaranteed to remains the same after `myhttp:start()`.
+    t.assert_equals(myhttp.tcp_server_f, tcp_server,
+        'tcp_server_f field was not changed after :start()')
+
+    myhttp:stop()
+    t.assert_equals(is_listening, false, 'custom socket was closed')
```
ligurio, line 49: I would also check that TCP port accepts connections after start() and TCP port refuse connections after stop(). Check with `is_listening` is good but not enough for reliable test.

ligurio (changes requested):
Thanks for the patch! Please see my comment regarding test.

test/unit/http_custom_socket_test.lua
Totktonada: Our stub socket replacement don't listen on a TCP port at all and it is enough to verify that the http server actually use the replacement. Is there a reason to make the test more complicated?

a1div0:
I have implemented a real connection test using the settings from `test.helpers`. Please review.

test/unit/http_custom_socket_test.lua
ligurio: > Our stub socket replacement don't listen on a TCP port at all and it is enough to verify that the http server actually use the replacement.

In such case, we don't need to check the state of the socket.

Totktonada: Reverted back to the test with stub socket.

Totktonada:
According to results of the discussion above I reverted it to the test with stub socket.

http/server.lua
```diff
@@ -1323,7 +1320,9 @@ local exports = {
             helper  = set_helper,
             hook    = set_hook,
             url_for = url_for_httpd,
-            listen  = tcp_server_create,
+
+            -- Exposed to make it replaceable by a user.
+            tcp_server_f = socket.tcp_server,
```
Totktonada, line 1325: Reminder: discuss this API with @ligurio and either squash or discard.

ligurio: Discussed orally. I would prefer API suggested by you.

ligurio (approved):
LGTM (after squashing commits)

a1div0:
Tell me please, what's the catch? I lost this think... What needs to be done to complete this pull request?

Totktonada:
No worry, I'll finish it.

http/server.lua
Totktonada: Squashed commits.

Totktonada (approved):

a1div0:
Thanks a lot!

PR #164, opened 2022-07-06
Artembo/packaging

artembo:


rpm/prebuild.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
Totktonada, line 1: How about `set -e`?

Totktonada: And `-o pipefail`.

rpm/prebuild.sh
```diff
@@ -0,0 +1,3 @@
...
+
+curl -LsSf https://www.tarantool.io/release/2/installer.sh | sudo bash
```
Totktonada, line 3: Shouldn't we use lowest supported tarantool version for building packages? It is 1.10 ATM.

If there are infrastructure problems on some OSes on 1.10, it is ok to use series-2 for them, but I would like this workaround being commented in the code.

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,31 @@
...
+        platform:
+          - { os: 'fedora', dist: '33' }
+
+    env:
+      OS: ${{ matrix.platform.os }}
+      DIST: ${{ matrix.platform.dist }}
+
+    steps:
+      - name: Clone the module
+        uses: actions/checkout@v2
```
Totktonada, line 22: Use `fetch-depth: 0` for correct `git describe` (and so correct package version). BTW, you can use newer checkout action version.

Totktonada: Also take care to tags deannotation: add `git fetch --tags -f`.

See the following links for details:

* https://github.com/tarantool/tarantool/blob/15b823f0ee3eb142c36564fa3973993c1ceba5b9/.github/actions/environment/action.yml#L52-L74
* https://github.com/actions/checkout/issues/290

artembo:
@Totktonada many thanks for the help with `rpm`, building packages with your suggestion works flawlessly.

Another issue is in `deb` package. There's an error during the build because of the absent `luatest`. If I disable testing in `debian/rules`, the build finishes successfully.

```diff
--- a/debian/rules
+++ b/debian/rules
@@ -2,7 +2,6 @@
 
 DEB_CMAKE_EXTRA_FLAGS := -DCMAKE_INSTALL_LIBDIR=lib/$(DEB_HOST_MULTIARCH) \
                        -DCMAKE_BUILD_TYPE=RelWithDebInfo
-DEB_MAKE_CHECK_TARGET := luatest
 
 include /usr/share/cdbs/1/rules/debhelper.mk
 include /usr/share/cdbs/1/class/cmake.mk
```

Following the approach for `rpm` .spec, can I get rid of `luatest` testing during the build?

rpm/prebuild.sh
artembo: I've changed installed Tarantool version to 1.10. Using the lowest Tarantool version seems the good practice.

.github/workflows/packaging.yml
Totktonada: Seems resolved.

artembo: yep, thanks for your help

Totktonada:
> Following the approach for `rpm` .spec, can I get rid of `luatest` testing during the build?

At least until we'll package `luatest`, it is okay.

artembo:
See dry run of building packages by introduced workflow [here](https://github.com/tarantool/http/actions/runs/2657157107)

rpm/tarantool-http.spec
```diff
@@ -20,14 +20,19 @@ This package provides a HTTP server for Tarantool.
 %setup -q -n %{name}-%{version}
 
 %build
-%cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo
-make %{?_smp_mflags}
-
-%check
-make %{?_smp_mflags} luatest
```
Totktonada, line 27: Nit: This change is not described/motivated anyhow in the commit message.

debian/prebuild.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+curl -LsSf https://www.tarantool.io/release/1.10/installer.sh | sudo bash
```
Totktonada, line 5: Nit: The change is titled 'deb: disable running tests on build' and it is a bit unexpected to see this addition here. Please, rewrite the commit title or split the changes.

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,89 @@
...
+        with:
+          fetch-depth: 0
```
Totktonada, line 49: Nit: It would be good to have a short comment why it is necessary.

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,89 @@
...
+          RWS_URL_PART: https://rws.tarantool.org/tarantool-modules
+          RWS_AUTH: ${{ secrets.RWS_AUTH }}
+          PRODUCT_NAME: tarantool-http
+        run: |
+          CURL_CMD="curl -LfsS \
+            -X PUT ${RWS_URL_PART}/${OS}/${DIST} \
+            -u ${RWS_AUTH} \
+            -F product=${PRODUCT_NAME}"
+
+          for f in $(ls -I '*build*' -I '*.changes' ./build); do
```
Totktonada, line 83: Ignore `*build*`, but include `./build`. Hm. Can you clarify this logic, please? Will not we ignore `./build`? Is it for `./build/build.log`? If so, maybe it is better to do `-I '*.log'`?

rpm/tarantool-http.spec
```diff
@@ -20,14 +20,19 @@ This package provides a HTTP server for Tarantool.
 %setup -q -n %{name}-%{version}
```
Totktonada, line 20: We should bump epoch due to #134. If it is solved somehow, I don't see how. Please, clarify.

debian/rules
```diff
@@ -2,7 +2,6 @@
```
Totktonada, line 2: We should bump epoch due to #134. If it is solved somehow, I don't see how. Please, clarify.

debian/prebuild.sh
artembo: Moved to the separate commit

.github/workflows/packaging.yml
artembo: done

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,89 @@
...
+        # setup. To avoid of it, let's fetch it manually, to be sure
+        # that all tags will exist always.
+        run: git fetch --tags -f
+
+      - name: Create packages
+        run: ./packpack/packpack
+
+      - name: Deploy packages
+        # We need this step to run only on push with tag.
+        if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') }}
```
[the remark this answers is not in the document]
artembo: I reckon we should. I've extended the list of triggers to:

```yml
on:
  pull_request:
  workflow_dispatch:
  push:
    tags:
      - '*'
```
After that we'll have an opportunity to test package building on each commit to any branch if the PR was created. `workflow_dispatch` will allow to trigger building manually. Anyway, publishing of the package will start only for pushing tags.

.github/workflows/packaging.yml
artembo: I see lots of stuff named using *build*. Probably, it's a good way to use `ls` in this case.
```bash
build
├── build.log
├── prebuild.sh
├── tarantool-http-1.2.0.10-1.el7.src.rpm
├── tarantool-http-1.2.0.10-1.el7.x86_64.rpm
├── tarantool-http-dbgsym_1.2.0.10-1_amd64.ddeb
├── tarantool-http-debuginfo-1.2.0.10-1.el7.x86_64.rpm
├── tarantool-http_1.2.0.10-1.debian.tar.xz
├── tarantool-http_1.2.0.10-1.dsc
├── tarantool-http_1.2.0.10-1_amd64.build
├── tarantool-http_1.2.0.10-1_amd64.buildinfo
├── tarantool-http_1.2.0.10-1_amd64.changes
├── tarantool-http_1.2.0.10-1_amd64.deb
└── tarantool-http_1.2.0.10.orig.tar.xz -> tarantool-http-1.2.0.10.tar.xz
```
If I didn't get you, please, explain again.

rpm/tarantool-http.spec
artembo: Epoch was already bumped [here](https://github.com/tarantool/http/commit/4db9aa3a16a767fd82b2125984af16c262b3ec6b#diff-c0279f62819ca56bac5a167034ae765318b020d36c8b151f8d999ba064c5ee39R4) and as we tested, the proper version of the package is installed.

rpm/tarantool-http.spec
artembo: done

artembo:
@Totktonada please review updated PR with resolving conversations if you are satisfied by the changes or answers.

debian/rules
artembo: Epoch for debian was already bumped [here](https://github.com/tarantool/http/commit/4db9aa3a16a767fd82b2125984af16c262b3ec6b#diff-c0279f62819ca56bac5a167034ae765318b020d36c8b151f8d999ba064c5ee39R4)

Totktonada: AFAIR, packpack replaces `Version` field in `debian/control` entirely. I guess we should teach packpack to replace just what is after the colon.

Am I wrong here?

artembo: I did some checks and what I get:

first, I build `tarantool-http` package using `packpack`

```bash
OS=ubuntu DIST=focal  ./packpack/packpack
```
then I run docker container to check which version of `tarantool-http` is installed

```bash
docker run --rm -it -v $(pwd)/build:/build ubuntu:focal bash
...
root@a3b036db23df:/# apt update && apt install -y curl
root@a3b036db23df:/# curl -LsSf https://www.tarantool.io/release/1.10/installer.sh |  bash
root@a3b036db23df:/# apt install -y tarantool tarantool-dev
root@a3b036db23df:/# cd build/
root@a3b036db23df:/build# dpkg -i tarantool-http_1.2.0.11-1_amd64.deb 
root@a3b036db23df:/build# apt-cache policy tarantool-http
tarantool-http:
  Installed: 1.2.0.11-1
  Candidate: 1:1.1.0.0-1
  Version table:
     1:1.1.0.0-1 500
        500 https://download.tarantool.org/tarantool/release/1.10/ubuntu focal/main amd64 Packages
        500 https://download.tarantool.org/tarantool/modules/ubuntu focal/main amd64 Packages
 *** 1.2.0.11-1 100
        100 /var/lib/dpkg/status
```

Okay, trying to build and install the same, but with shallow git clone. Whay I get after the same operations as I did above:

```bash
root@4ca7d7f7c3d5:/build# apt-cache policy tarantool-http
tarantool-http:
  Installed: 0.0.1-1
...
```

The version of the package in it's spec is not affecting on a build using `packpack` as it gets version from `git describe`. I see a problem in the current uploaded version in the Tarantool, which contains epoch version.

Somehow `tarantool-http` version 1:1.1.0 was uploaded into repository (prod) and it has bigger priority, so I copy it into dev repo as well.

Let's try to upload tarantool-http to the test repo and see which module will be installed.

Current repo list:
```bash
cat /etc/apt/sources.list.d/tarantool_2.list
deb https://download.tarantool.org/tarantool/release/series-2/ubuntu/ focal main
deb https://hb.bizmrg.com/tarantool-dev/restored/release/modules/ubuntu/ focal main
```

Check package candidate:
```bash
root@7f0778511fce:/# apt-cache policy tarantool-http
tarantool-http:
  Installed: (none)
  Candidate: 1:1.1.0.0-1
  Version table:
     1:1.1.0.0-1 500
        500 https://hb.bizmrg.com/tarantool-dev/restored/release/modules/ubuntu focal/main amd64 Packages
     1.2.0.10-1 500
        500 https://hb.bizmrg.com/tarantool-dev/restored/release/modules/ubuntu focal/main amd64 Packages
```

Alexander, please tell me the solution to this problem, I'm stuck.

.github/workflows/packaging.yml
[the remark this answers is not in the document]
Totktonada: Ah, I see, it is on any pull request: from the base repository branch or from a fork. But how about `master`? I think we should run the testing on pushing to `master` as well.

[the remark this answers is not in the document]
artembo: I think it's better to tag more often, because we don't use this practice widely, for example, we don't publish `Tarantool` on each commit to `master`. If you insist, we can push packages on each commit to `master`. In the suggested configuration we have only testing builds on each push.

.github/workflows/packaging.yml
Totktonada: `-I build.log -I prebuild.sh -I '*.build' -I '*.buildinfo'` would be self-descriptive IMHO. `*build*` looks cryptic.

artembo: As we need only `deb` and `rpm` files, probably it's better to do like this:
```bash
for f in $(cd build; ls *{rpm,deb}); do
  CURL_CMD+=" -F $(basename ${f})=@./build/${f}"
done
```

.github/workflows/packaging.yml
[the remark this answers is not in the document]
artembo: I agree, enabled testing on pushing to `master`

.github/workflows/packaging.yml
Totktonada: 1. I would prefer `*.foo` over `*foo` -- less space for mistakes.
2. Will RWS accept `*.ddeb` (it is matched by the `*deb` glob)? See https://github.com/tarantool/rws/issues/13.
3. Shouldn't we store `*.dsc` files? RWS accepts them at least: https://github.com/tarantool/mkrepo/pull/61.
4. Here I don't know, just question: shouldn't we store debian source tarballs (`*.debian.tar.[gx]z`, `*.orig.tar.[gx]z`)?

debian/rules
Totktonada: packpack replaces the version from a repository to `git describe` based one (or to one provided in the `VERSION` environment variable) as for RPM as well as for Deb packages. AFAIS, the relevant code for Debian/Ubuntu is [here](https://github.com/packpack/packpack/blob/194f898750cbbee6dd35255fe8aa7ec878eb3836/pack/deb.mk#L106-L109).

There are different possible ways to teach packpack to accept our epoch:

1. Take care to EPOCH environment variable (we should also support it for RPMs, if we'll do it for Debs).
2. Extract `Version: (.*)` from `debian/control` and, if the verison has epoch, take care to this epoch value.
3. Both 1 and 2.

I like the second more than the first, because it allows to write the epoch into the repository once (like we do in 4db9aa3a16a767fd82b2125984af16c262b3ec6b) and call `OS=<...> DIST=<...> /path/to/packpack` as usual. It is most convenient in our case. However the first one also may be useful.

BTW, there is related issue in packpack: https://github.com/packpack/packpack/issues/79.

.github/workflows/packaging.yml
LeonidVas: I’ll give you my 2 cents.
- we do not publish `*.ddeb` packages and I see no reason to start doing that. 
- we also publish `tar.xz`, `debian.tar.xz`, `orig.tar.xz` and `dsc` files. I suggest do it the same way we publish other packages, with the same set of files ([example](https://rws.tarantool.org/release/modules/ubuntu/pool/groovy/main/t/tarantool-queue))

artembo: Okay, thanks! This solution seems to be complete and descriptive:
```bash
for f in $(cd build; ls *.(deb|rpm|dsc|tar.xz)); do
      CURL_CMD+=" -F $(basename ${f})=@./build/${f}"
done
```

debian/rules
Totktonada: It seems, it'll be fixed on the packpack side. Closing the discussion so.

.github/workflows/packaging.yml
artembo: @Totktonada, what do you think about this solution? `shopt -s nullglob` allows to ignore not-matched patterns, and the cycle looks clear. The `woking-directory` is set in the job manifest.
```bash
shopt -s nullglob
for f in *.deb *.rpm *.dsc *.tar.xz; do
    CURL_CMD+=" -F $(basename ${f})=@${f}"
done
```

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,93 @@
...
+          - { os: 'fedora', dist: '32' }
+          - { os: 'fedora', dist: '33' }
+          - { os: 'fedora', dist: '34' }
+          - { os: 'fedora', dist: '35' }
+          - { os: 'fedora', dist: '36' }
+          - { os: 'ubuntu', dist: 'xenial' }
+          - { os: 'ubuntu', dist: 'bionic' }
+          - { os: 'ubuntu', dist: 'focal' }
+          - { os: 'ubuntu', dist: 'groovy' }
+          - { os: 'ubuntu', dist: 'hirsute' }
```
Totktonada, line 36: CI is red here.

.github/workflows/packaging.yml
Totktonada: I'm ok. The only comment: it is not a mistake that I said `.tar.[gx]z`, not just `.tar.xz`. I don't remember, when precisely it was changed, but I remember this PR: https://github.com/tarantool/mkrepo/pull/15.

Totktonada (approved):
I'm ok with current state of the PR aside of two remaining comments. Let's fix them and take my approve.

Please, proceed with Leonid for further steps.

.github/workflows/packaging.yml
LeonidVas: image
https://github.com/tarantool/queue/commit/8f5cc970aba0d1ab5b4858fab767ad297275daee#diff-b262e6f808c6e8d409cdeeeec179bcf85a312d8ad9e16b8c7861dbe538b62b0b

.github/workflows/packaging.yml
artembo: added *.tar.gz

.github/workflows/packaging.yml
artembo: These fails caused by disabling Ubuntu repositories of EOL hisrute and impish. @0x501D, can I remove these OS versions from the workflow?

0x501D: Yes, I see no reason to run CI on EOL distributions

LeonidVas (approved):
LGTM

0x501D (approved):
LGTM

.github/workflows/packaging.yml
```diff
@@ -0,0 +1,91 @@
...
+          - { os: 'fedora', dist: '31' }
+          - { os: 'fedora', dist: '32' }
+          - { os: 'fedora', dist: '33' }
+          - { os: 'fedora', dist: '34' }
+          - { os: 'fedora', dist: '35' }
+          - { os: 'fedora', dist: '36' }
+          - { os: 'ubuntu', dist: 'xenial' }
+          - { os: 'ubuntu', dist: 'bionic' }
+          - { os: 'ubuntu', dist: 'focal' }
+          - { os: 'ubuntu', dist: 'groovy' }
```
NickVolynkin, line 35: Ubuntu Groovy 20.10 is EOL

Issue #171, opened 2022-09-15
Failed to parse request: invalid uri


eugenepaniot:
https://github.com/tarantool/http/blob/606fa09bf4e8e4bd911e4d78a9d3fd3ee5042cdb/http/server.lua#L738

Tarantool suggests the url request `/some/valid/path./` as invalid, but it should not.

```
httpd:route({ path = '/some/valid/*path', method = 'GET', public = true }, http_middleware.v1(handle_get_path, http_collector) )
```

I need to get request `path` as is.
Thank you.

**How to reproduce:**

```
$ curl -v 'http://127.1:8081/some/valid/path./'

*   Trying 127.0.0.1:8081...
* Connected to 127.0.0.1 (127.0.0.1) port 8081 (#0)
> GET /some/valid/path./ HTTP/1.1
> Host: 127.0.0.1:8081
> User-Agent: curl/7.84.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
* HTTP 1.0, assume close after body
< HTTP/1.0 400 Bad request
<
* Closing connection 0
invalid uri
```

Totktonada:
NB: It comes from ff01a9400f30633d135f22f48e2e110923d6fe26.

Totktonada:
So it was implemented for the http client (now an http client is built into tarantool itself). And it is anyway a strange thing. It is valid URI with relative path. I don't see a proper reason to ban it.

0x501D:
Does the http client handle "./" and "../" as per rfc3986 clause 5.2.4. "Remove Dot Segments"?

PR #174, opened 2022-11-07
Do not treat ./ as invalid uri

eugenepaniot:
Do not treat ./ as invalid uri.
Fixes https://github.com/tarantool/http/issues/171.

```
$ .rocks/bin/luatest -v test/
...
=========================================================
Ran 56 tests in 0.648 seconds, 56 succeeded, 0 failed
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Add versioning support.
 
+### Fixed
+
+- Allow dot in path segment.
+
 ## [1.4.0] - 2022-12-30
 
 ### Added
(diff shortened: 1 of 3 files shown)


0x501D:
This patch resulted in an undefined instruction in the case of "../" processing. The patch should handle both cases: "./" and "../" according to rfc3986 clause 5.2.4. @Totktonada , correct me if I'm wrong.

Totktonada:
@0x501D I would allow those paths as a bugfix and consider a path resolution as a separate feature. I'm not sure, whether the standard obligates us to do the path resolution, but anyway it is its own task.

eugenepaniot:
JFYI nginx passes it as is.

Totktonada (approved):

PR #176, opened 2022-11-15
cd: fix packaging for CentOS7

0x501D:
Fix the problem with sudoers for CentOS7.

For more details see:
https://github.com/packpack/packpack/issues/145


LeonidVas (approved):
LGTM

PR #180, opened 2022-11-29
Fix description for req:request_line method

### LINKED ISSUE #128: req:request_line() doesn't return with the body
Here is the test body
`{"user":{"email":"vmi@vmi.hu","password":"a1"}}`
req:request_line() returns with
`POST /ugyfelkapu-api/v1/login HTTP/1.1`
but the description says
> returns the request body.

I think something isn't good maybe the code or the desc. 
I need to get back the body and as a string.

lowitea:
closes #128


README.md
```diff
@@ -249,7 +249,7 @@ end
 * `req.peer` - a Lua table with information about the remote peer
   (like `socket:peer()`).
 * `tostring(req)` - returns a string representation of the request.
-* `req:request_line()` - returns the request body.
+* `req:request_line()` - returns a first line of the http request packet (for example, `PUT /path HTTP/1.1`).
```
Totktonada, line 252: Nit: I would replace 'the http request packet' with just 'the http request', because 'packet' is a TCP/IP level term or at least feels so.

Totktonada (approved):

README.md
Totktonada: Nice! Excuse me for nits, but, please, squash the commits and force-push the resulting one.

Totktonada: Thanks!

0x501D (approved):

PR #197, opened 2024-09-05
roles: introduce role for http servers

### LINKED ISSUE #196: Introduce server setup through Tarantool 3 role
[`metrics-export-role 0.1.0`](https://github.com/tarantool/metrics-export-role/) had introduced a way to set up metrics HTTP server through Tarantool 3 configuration:
```yaml
roles_cfg:
  roles.metrics-export:
    http:
    - listen: 8081
      endpoints:
      - path: /metrics
        format: json
      - path: /metrics/prometheus/
        format: prometheus
    - listen: '127.0.0.1:8082'
      endpoints:
      - path: /metrics/
        format: json
```

The next step is to introduce some configurable HTTP server that can be reused by several roles (like with Tarantool Cartridge Web UI HTTP server). It can go like this:

```yaml
roles_cfg:
  roles.http:
    default_server:
    - listen: 8081
    additional_server:
    - listen: '127.0.0.1:8082'
  roles.metrics-export:
    http:
    - server: default_server
      endpoints:
      - path: /metrics
        format: json
      - path: /metrics/prometheus/
        format: prometheus
    - server: additional_server
      endpoints:
      - path: /metrics/
        format: json
  roles.tcf-worker:
    http:
    - server: default_server
```

Such approach will also allow to set up server port on instance level and metrics/other HTTP users on global level instead of copypasting all of it to the instance level.

So HTTP role should support
- creating servers with some unique identifier,
- API to get server object by its ID for other roles.

It would also be nice to have some "get_default_server" handle so it can be used like this:
```yaml
roles_cfg:
  roles.http:
    default:
    - listen: 8081
  roles.metrics-export:
    http:
      endpoints:
      - path: /metrics
        format: json
      - path: /metrics/prometheus/
        format: prometheus
```

Default server can be a "first server created with role" or "server with `default` key". This case should be useful since "single HTTP server for administration and monitoring" is a well-spread case from Tarantool Cartridge applications.

(Configuration structure proposed here is an example and not a strict requirement.)

DerekBum:
This patch adds `roles.http_server`. This role allows to configurate one or more HTTP servers. Those servers could be reused by several other roles.

Servers could be accessed by their names (from the config).

`get_server(name)` method returns a server by its name. If `nil` is passed, default server is returned. The server is  default, if it has `DEFAULT_SERVER_NAME` (`"default"`) as a name.

This patch also adds a `httpd:delete(name)` method to delete named routes.

Closes #196

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,17 +10,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tarantool: ['1.10', '2.5', '2.6', '2.7', '2.8']
+        tarantool: ['1.10', '2.10', '2.11', '3.1', '3.2']
         coveralls: [false]
         include:
-          - tarantool: '2.10'
+          - tarantool: '2.11'
             coveralls: true
     runs-on: [ubuntu-20.04]
     steps:
       - uses: actions/checkout@master
-      - uses: tarantool/setup-tarantool@v1
+      - uses: tarantool/setup-tarantool@v3
         with:
           tarantool-version: ${{ matrix.tarantool }}
+
+      - name: Prepare the repo
+        run: curl -L https://tarantool.io/release/2/installer.sh | bash
+        env:
+          DEBIAN_FRONTEND: noninteractive
+
+      - name: Install tt cli
+        run: sudo apt install -y tt=2.4.0
+        env:
+          DEBIAN_FRONTEND: noninteractive
 
       - name: Cache rocks
         uses: actions/cache@v2
@@ -40,8 +50,9 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Run tests and code coverage analysis
-        run: make -C build coverage
+      - name: Run tests without code coverage analysis
+        run: make -C build luatest
+        if: matrix.coveralls != true
 
       - name: Send code coverage to coveralls.io
         run: make -C build coveralls
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -9,6 +9,8 @@ ignore = {
     "143/string",
     -- Accessing an undefined field of a global variable <table>.
     "143/table",
+    -- Accessing an undefined field of a global variable <package>.
+    "143/package",
     -- Unused argument <self>.
     "212/self",
     -- Redefining a local variable.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - `roles.httpd` role to configure one or more HTTP servers (#196)
+- `httpd:delete(name)` method to delete named routes (#197)
 
 ## [1.5.0] - 2023-03-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Fixed
+
 - Fixed request crash with empty body and unexpected header Content-Type (#189)
+
+### Added
+
+- `roles.httpd` role to configure one or more HTTP servers (#196)
 
 ## [1.5.0] - 2023-03-29
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Wall -Wextra")
 string(RANDOM ALPHABET 0123456789 seed)
 
 add_subdirectory(http)
+add_subdirectory(roles)
 
 add_custom_target(luacheck
   COMMAND ${LUACHECK} ${PROJECT_SOURCE_DIR}
@@ -32,11 +33,17 @@ add_custom_target(luacheck
   COMMENT "Run luacheck"
 )
 
-add_custom_target(luatest
+add_custom_target(luatest-coverage
   COMMAND ${LUATEST} -v --coverage --shuffle all:${seed}
   BYPRODUCTS ${CODE_COVERAGE_STATS}
   WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-  COMMENT "Run regression tests"
+  COMMENT "Run regression tests with coverage"
+)
+
+add_custom_target(luatest
+  COMMAND ${LUATEST} -v --shuffle all:${seed}
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  COMMENT "Run regression tests without coverage"
 )
 
 add_custom_target(coverage
@@ -65,4 +72,4 @@ add_custom_target(coveralls
 
 set (LUA_PATH "LUA_PATH=${PROJECT_SOURCE_DIR}/?.lua\\;${PROJECT_SOURCE_DIR}/?/init.lua\\;\\;")
 set (LUA_SOURCE_DIR "LUA_SOURCE_DIR=${PROJECT_SOURCE_DIR}")
-set_target_properties(luatest PROPERTIES ENVIRONMENT "${LUA_PATH};${LUA_SOURCE_DIR}")
+set_target_properties(luatest-coverage PROPERTIES ENVIRONMENT "${LUA_PATH};${LUA_SOURCE_DIR}")
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ http v2 revert and decisions regarding each reverted commit see
   * [before\_dispatch(httpd, req)](#before_dispatchhttpd-req)
   * [after\_dispatch(cx, resp)](#after_dispatchcx-resp)
 * [Using a special socket](#using-a-special-socket)
+* [Roles](#roles)
+  * [roles.httpd](#roleshttpd)
 * [See also](#see-also)
 
 ## Prerequisites
@@ -501,6 +503,114 @@ server:start()
 ```
 
 [socket_ref]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/#socket-tcp-server
+
+## Roles
+
+Tarantool 3 roles could be accessed from this project.
+
+### `roles.httpd`
+
+It allows configuring one or more HTTP servers. Those servers could be reused
+by several other roles.
+
+Example of the configuration:
+
+```yaml
+roles_cfg:
+  roles.httpd:
+    default:
+    - listen: 8081
+    additional:
+    - listen: '127.0.0.1:8082'
+```
+
+Server address should be provided either as a URI or as a single port
+(in this case, `0.0.0.0` address is used).
+
+User can access every working HTTP server from the configuration by name,
+using `require('roles.httpd').get_server(name)` method.
+If the `name` argument is `nil`, the default server is returned
+(its name should be equal to constant
+`require('roles.httpd').DEFAULT_SERVER_NAME`, which is `"default"`).
+
+Let's look at the example of using this role. Consider a new role
+`roles/hello_world.lua`:
+```lua
+local M = { dependencies = { 'roles.httpd' } }
+local server = {}
+
+M.validate = function(conf)
+    if conf == nil or conf.httpd == nil then
+        error("httpd must be set")
+    end
+    local server = require('roles.httpd').get_server(conf.httpd)
+    if server == nil then
+        error("the httpd server " .. conf.httpd .. " not found")
+    end
+end
+
+M.apply = function(conf)
+    server = require('roles.httpd').get_server(conf.httpd)
+
+    server:route({
+        path = '/hello/world',
+        name = 'greeting',
+    }, function(tx)
+        return tx:render({text = 'Hello, world!'})
+    end)
+end
+
+M.stop = function()
+    server:delete('greeting')
+end
+
+return M
+```
+
+This role accepts a server by name from a config and creates a route to return 
+`Hello, world!` to every request by this route.
+
+Then we need to write a simple config to start the Tarantool instance via
+`tt`:
+```yaml
+app:
+  file: 'myapp.lua'
+
+groups:
+  group001:
+    replicasets:
+      replicaset001:
+        roles: [roles.httpd, roles.hello_world]
+        roles_cfg:
+          roles.httpd:
+            default:
+              listen: 8081
+            additional:
+              listen: '127.0.0.1:8082'
+          roles.hello_world:
+            httpd: 'additional'
+        instances:
+          instance001:
+            iproto:
+              listen:
+                - uri: '127.0.0.1:3301'
+```
+
+Next step, we need to start this instance using `tt start`:
+```bash
+$ tt start
+   • Starting an instance [app:instance001]...
+$ tt status
+ INSTANCE         STATUS   PID      MODE 
+ app:instance001  RUNNING  2499387  RW
+```
+
+And then, we can get the greeting by running a simple curl command from a
+terminal:
+```bash
+$ curl http://127.0.0.1:8082/hello/world
+Hello, world!
+```
 
 ## See also
 
--- a/README.md
+++ b/README.md
@@ -184,6 +184,13 @@ httpd:route({ path = '/objects', method = 'GET' }, handle3)
 ...
 ```
 
+To delete a named route, use `delete()` method of the `httpd` object:
+
+```lua
+httpd:route({ path = '/path/to', name = 'route' }, 'controller#action')
+httpd:delete('route')
+```
+
 The first argument for `route()` is a Lua table with one or more keys:
 
 * `file` - a template file name (can be relative to.
--- a/deps.sh
+++ b/deps.sh
@@ -5,13 +5,14 @@
 set -e
 
 # Test dependencies:
-tarantoolctl rocks install luatest 0.5.7
-tarantoolctl rocks install luacheck 0.25.0
-tarantoolctl rocks install luacov 0.13.0
-tarantoolctl rocks install luafilesystem 1.7.0-2
+# Could be replaced with luatest >= 1.1.0 after a release.
+tt rocks install luatest
+tt rocks install luacheck 0.25.0
+tt rocks install luacov 0.13.0
+tt rocks install luafilesystem 1.7.0-2
 
 # cluacov, luacov-coveralls and dependencies
-tarantoolctl rocks install luacov-coveralls 0.2.3-1 --server=https://luarocks.org
-tarantoolctl rocks install cluacov 0.1.2-1 --server=https://luarocks.org
+tt rocks install luacov-coveralls 0.2.3-1 --server=https://luarocks.org
+tt rocks install cluacov 0.1.2-1 --server=https://luarocks.org
 
-tarantoolctl rocks make
+tt rocks make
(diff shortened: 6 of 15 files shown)


oleg-jukovec:
Let's also add a public API call for this functionality in the scope of the task, so that there is no need to access internal structures directly:

https://github.com/tarantool/metrics-export-role/blob/36130b73c0a5975ff26f0c26f78ee7a368c1207b/roles/metrics-export.lua#L26-L37

http-scm-1.rockspec
```diff
@@ -31,6 +31,7 @@ build = {
         ['http.version'] = 'http/version.lua',
         ['http.mime_types'] = 'http/mime_types.lua',
         ['http.codes'] = 'http/codes.lua',
+        ['roles.http'] = 'roles/http.lua',
```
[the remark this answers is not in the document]
DerekBum: It does not seem like it (or maybe I don't quite understand where to look). In `rpm` folder nothing needs an update.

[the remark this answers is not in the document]
oleg-jukovec: https://github.com/tarantool/http/blob/0b471d8ce9ef212911c1dba3b6831c142f540cfa/CMakeLists.txt#L27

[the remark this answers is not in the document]
DerekBum: Should I create an own CMakeList for the `roles` subdirectory, or simply install it from the `http`?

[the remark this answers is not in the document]
oleg-jukovec: For me it looks like this. You could check result packages with `packpack/packpack`.

DerekBum:
Updated the PR according to the comments.

http/server.lua
```diff
@@ -1357,6 +1374,7 @@ local exports = {
 
             -- methods
             route   = add_route,
+            delete  = delete_route,
```
oleg-jukovec, line 1377: We need a test for the API.

README.md
```diff
@@ -502,6 +511,38 @@ server:start()
...
+Server address should be provided either as a URI or as a single port
+(in this case, `0.0.0.0` address is used).
+
+User can access every working HTTP server from the configuration by name,
+using `get_server(name)` method.
+If the `name` argument is `nil`, the default server is returned
+(its name should be equal to constant `DEFAULT_SERVER_NAME`).
+
+For an example of using this role, refer to
+[this integration test](./test/integration/http_server_role_test.lua).
```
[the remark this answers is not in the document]
DerekBum: Right now no other project supports this role. And to show an actual example, AFAIU, we either need to show how to support this role from the other role (which is done in `./test/integration/http_server_role_test.lua` via custom mock role), or just write some pulled from thin air example directly in `README`.

For instance, the second way will look like this:
`To show an example on how to use this role with other roles, lets consider role *my_role*, which uses HTTP servers for its own purposes. To configure the connection to each server, the config expects the name of the server and some endpoints to connect. In this example, config file may look like this:`
```yml
roles_cfg:
  roles.http_server:
    default:
    - listen: 8081
    additional:
    - listen: '127.0.0.1:8082'
  roles.my-role:
    - endpoints:
      - path: /metrics
      - path: /metrics/prometheus/
        format: prometheus
    - server: additional
      endpoints:
      - path: /metrics/
```
`Notice that the first connection will use the default server, since no name is provided.`

Or something like that. Will this be ok to put as the doc?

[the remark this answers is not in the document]
oleg-jukovec: I would like to see a code of minimal role that output "Hello world" + an example of the configuration for the role:
```yaml
roles_cfg:
  roles.http_server:
    default:
    - listen: 8081
    additional:
    - listen: '127.0.0.1:8082'
  roles.my-role:
    - httpd: additional
```

.luacheckrc
```diff
@@ -1,5 +1,5 @@
 std = "luajit"
-globals = {"box", "_TARANTOOL", "tonumber64", "utf8"}
+globals = {"box", "_TARANTOOL", "tonumber64", "utf8", "checks"}
```
DifferentialOrange, line 2: Let's use `local checks = require('checks')` instead

DerekBum:
CI is finally green, all comments are fixed.

.github/workflows/test.yml
```diff
@@ -10,18 +10,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tarantool: ['1.10', '2.5', '2.6', '2.7', '2.8']
+        tarantool: ['1.10', '2.5', '2.6', '2.7', '2.8', '3.2']
```
oleg-jukovec, line 13: Let's remove the outdated versions:
```suggestion
        tarantool: ['1.10', '2.10', '2.11', '3.1', '3.2']
```

CHANGELOG.md
```diff
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Fixed
+
 - Fixed request crash with empty body and unexpected header Content-Type (#189)
 
+### Added
+
+- `httpd` role to configure one or more HTTP servers (#196)
```
oleg-jukovec, line 15:
```suggestion
- `roles.httpd` role to configure one or more HTTP servers (#196)
```

CMakeLists.txt
```diff
@@ -39,6 +40,12 @@ add_custom_target(luatest
   COMMENT "Run regression tests"
 )
 
+add_custom_target(luatest-no-coverage
```
oleg-jukovec, line 43: It's better to do the opposite.
`luatest` -> `luatest-coverage`
`luatest-no-coverage` -> `luatest`
```suggestion
add_custom_target(luatest
```

CMakeLists.txt
```diff
@@ -39,6 +40,12 @@ add_custom_target(luatest
...
+  COMMAND ${LUATEST} -v
```
oleg-jukovec, line 44: `shuffle` is not related to the coverage.
```suggestion
  COMMAND ${LUATEST} -v --shuffle all:${seed}
```

README.md
```diff
@@ -502,6 +511,91 @@ server:start()
 
 [socket_ref]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/#socket-tcp-server
 
+## Roles
+
+New roles could be accessed from this project.
+
+### `httpd` role
```
oleg-jukovec, line 518:
```suggestion
### `roles.httpd` role
```

README.md
```diff
@@ -502,6 +511,91 @@ server:start()
...
+    - listen: 8081
+    additional:
+    - listen: '127.0.0.1:8082'
+```
+
+Server address should be provided either as a URI or as a single port
+(in this case, `0.0.0.0` address is used).
+
+User can access every working HTTP server from the configuration by name,
+using `get_server(name)` method.
```
oleg-jukovec, line 538:
```suggestion
using `require('roles.httpd').get_server(name)` method.
```

README.md
```diff
@@ -502,6 +511,91 @@ server:start()
...
+If the `name` argument is `nil`, the default server is returned
+(its name should be equal to constant `DEFAULT_SERVER_NAME`).
```
oleg-jukovec, line 540: And what is a value of this constant?

README.md
```diff
@@ -502,6 +511,91 @@ server:start()
...
+        path = '/hello/world'
+    }, function(tx)
+        return tx:render({text = 'Hello, world!'})
+    end)
+end
+
+M.get = function()
+    local server_uri = ('http://%s:%s'):format(server.host, server.port)
+    return http_client.get(server_uri .. '/hello/world')
+end
```
oleg-jukovec, line 562: Let's remove the code and just make a request in an example, using `curl`.

README.md
```diff
@@ -502,6 +511,91 @@ server:start()
...
+Tarantool instance:
+```lua
+local httpd_role = require('roles.httpd')
+local hello_role = require('roles.hello_world')
+
+httpd_role.apply(config.roles_cfg.httpd)
+hello_role.apply(config.roles_cfg.hello_world, httpd_role)
+
+print(hello_role.get().body)
+```
```
oleg-jukovec, line 597: Nobody starts Tarantool like that, we do it in such way for testing. It's better to replace the example with a real cluster configuration file + `tt start`. See examples in Tarantool's docs: https://github.com/tarantool/doc/tree/latest/doc/code_snippets/snippets .

deps.sh
```diff
@@ -5,13 +5,26 @@
 set -e
 
 # Test dependencies:
-tarantoolctl rocks install luatest 0.5.7
-tarantoolctl rocks install luacheck 0.25.0
-tarantoolctl rocks install luacov 0.13.0
-tarantoolctl rocks install luafilesystem 1.7.0-2
+tt rocks install luatest
```
oleg-jukovec, line 8: We could specify a `luatest` version here too.
```suggestion
tt rocks install luatest
```

deps.sh
```diff
@@ -5,13 +5,26 @@
...
+tt rocks install luacheck 0.25.0
+tt rocks install luacov 0.13.0
 
 # cluacov, luacov-coveralls and dependencies
-tarantoolctl rocks install luacov-coveralls 0.2.3-1 --server=https://luarocks.org
-tarantoolctl rocks install cluacov 0.1.2-1 --server=https://luarocks.org
+tt rocks install https://raw.githubusercontent.com/luarocks/cluacov/master/cluacov-dev-1.rockspec
+tt rocks install https://luarocks.org/manifests/dhkolf/dkjson-2.8-1.rockspec
+tt rocks install https://raw.githubusercontent.com/keplerproject/luafilesystem/master/luafilesystem-scm-1.rockspec
+tt rocks install https://raw.githubusercontent.com/moteus/lua-path/master/rockspecs/lua-path-scm-0.rockspec
```
oleg-jukovec, line 16: Why are these dependencies now installed that way?

deps.sh
DerekBum: Both luatest 1.0.1 and 1.0.1-1 does not contain treegen, which is used in one of the tests.

README.md
DerekBum: `default` (this constant is also accessible from the role variable).

.github/workflows/test.yml
DerekBum: Don't we want to remove 1.10 completely then?

README.md
oleg-jukovec: Please, add it to the document. Configuration from the config should ideally be done without code.

.github/workflows/test.yml
oleg-jukovec: This is a debatable issue, but it seems to me that there should be testing on branch 1.x until it breaks. This way we will know that we are still working on 1.x.

deps.sh
oleg-jukovec: Please, add a comment like that:

```
Could be replaced with luatest >= 1.1.0 after a release.
```

It will help us do not forget the reason.

DerekBum:
Updated the PR according to comments.

README.md
```diff
@@ -502,6 +511,115 @@ server:start()
...
+    }, function(tx)
+        return tx:render({text = 'Hello, world!'})
+    end)
+end
+
+M.stop = function()
+    server:delete('greeting')
+end
+
+return M
```
oleg-jukovec, line 566:
```suggestion
local M = {}

M.validate = function(conf)
    if conf == nil or conf.httpd == nil then
        error("httpd must be set")
    end
    local server = require('roles.httpd').get_server(conf.httpd).httpd
    if server == nil then
        error("the httpd server " .. conf.httpd .. " not found")
    end
end

M.apply = function(conf)
    local server = require('roles.httpd').get_server(conf.httpd).httpd

    server:route({
        path = '/hello/world',
        name = 'greeting',
    }, function(tx)
        return tx:render({text = 'Hello, world!'})
    end)
end

M.stop = function()
    local server = require('roles.httpd').get_server(conf.httpd).httpd    
    server:delete('greeting')
end

return M
```

oleg-jukovec: It would be nice to add a dependency `roles.httpd` for the role too:
https://www.tarantool.io/en/doc/latest/platform/app/app_roles/#role-dependencies

README.md
```diff
@@ -502,6 +511,115 @@ server:start()
...
+
+In `myapp.lua` we need to write a script to initialize both roles:
+```lua
+local httpd_role = require('roles.httpd')
+local hello_role = require('roles.hello_world')
+
+httpd_role.apply(config.roles_cfg.httpd)
+hello_role.apply(config.roles_cfg.hello_world)
+```
+
```
oleg-jukovec, line 606: It could be removed, just config.yaml + tt start should be enough.

oleg-jukovec (approved):
Thank you for the patch, just few non-critical notes.

CHANGELOG.md
```diff
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Fixed
+
 - Fixed request crash with empty body and unexpected header Content-Type (#189)
 
+### Added
+
+- `roles.httpd` role to configure one or more HTTP servers (#196)
+- `httpd:delete(name)` method to delete named routes (#197)
```
DifferentialOrange, line 16: I have never seen anyone using `name` option for a route. Until now, I didn't know it even exists. As far as I understand, it actually works not only with named routes, but with unnamed ones as well by using path as a key?

README.md
```diff
@@ -502,6 +511,115 @@ server:start()
...
+Let's look at the example of using this role. Consider a new role
+`roles/hello_world.lua`:
+```lua
+local M = {}
+local server = {}
+
+M.validate = function()  end
+
+M.apply = function(conf)
+    server = require('roles.httpd').get_server(conf.httpd).httpd
```
DifferentialOrange, line 552: > get_server().httpd

Why don't we return `httpd` itself? If I got it right, the `server` consists of `httpd` and `routes`, while `routes` is some internal `role.httpd` thing that is not expected to be used or changed by external roles.

README.md
DifferentialOrange: Yes, `role.apply` is not designed to be called directly in applications/other roles. (We may do so in unit tests, but just for the sake of unit testing.)

.github/workflows/test.yml
```diff
@@ -40,8 +50,9 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Run tests and code coverage analysis
-        run: make -C build coverage
+      - name: Run tests without code coverage analysis
```
DifferentialOrange, line 53: So it seems that Tarantool 3.2 coverage still segfaults? What's about running coverage with 2.11?

README.md
```diff
@@ -502,6 +511,115 @@ server:start()
...
+using `require('roles.httpd').get_server(name)` method.
+If the `name` argument is `nil`, the default server is returned
+(its name should be equal to constant
+`require('roles.httpd').DEFAULT_SERVER_NAME`, which is `"default"`).
+
+Let's look at the example of using this role. Consider a new role
+`roles/hello_world.lua`:
+```lua
+local M = {}
+local server = {}
```
DifferentialOrange, line 547: The reason to cache it here is to do a `stop` cleanup?

roles/httpd.lua
```diff
@@ -0,0 +1,133 @@
+local checks = require('checks')
+local urilib = require("uri")
```
DifferentialOrange, line 2:
```suggestion
local urilib = require('uri')
```

test/integration/httpd_role_test.lua
```diff
@@ -0,0 +1,76 @@
...
+        },
+    }
+    local dir = treegen.prepare_directory({}, {})
+
+    local config_file = treegen.write_file(dir, 'config.yaml',
+        yaml.encode(config))
+    local opts = {config_file = config_file, chdir = dir}
+    g.server = server:new(fun.chain(opts, {alias = 'instance-001'}):tomap())
+
+    g.server:start()
```
DifferentialOrange, line 69: I think it should be done in `before_all`/`before_test` hook so you can clean it up in `after_all`/`after_each`.

test/integration/httpd_role_test.lua
```diff
@@ -0,0 +1,76 @@
...
+
+    httpd_role.apply(config.roles_cfg.httpd)
```
DifferentialOrange, line 71: Adding them to instance config section `roles = ['roles.http', 'test.mocks.mock_role']` should be enough.

CHANGELOG.md
DerekBum: No, we can't use just path. Routes are stored as `opts` (which is a table of many fields, some of them are filled inside of `add_route` function).
https://github.com/tarantool/http/blob/2e3dbd2935b3ecf8ce5e4eae62933762e0ee576a/http/server.lua#L1248-L1259

And those `opts` stored in the map with the `name` as a key. If there are no `name`, then we can't get `opts`.

README.md
DerekBum: Yes.

CHANGELOG.md
DifferentialOrange: Oh, I see that metrics-export-role sets the name as well
https://github.com/tarantool/metrics-export-role/blob/36130b73c0a5975ff26f0c26f78ee7a368c1207b/roles/metrics-export.lua#L265

oleg-jukovec: The only problem is to create a nice API for the call. Maybe:
```
http:delete(path, name)
http:delete(path)
http:delete(nil, name)
```

DerekBum:
Updated everything according to comments.

README.md
```diff
@@ -502,6 +511,114 @@ server:start()
 
 [socket_ref]: https://www.tarantool.io/en/doc/latest/reference/reference_lua/socket/#socket-tcp-server
 
+## Roles
+
+New roles could be accessed from this project.
```
oleg-jukovec, line 516:
```suggestion
Tarantool 3 roles could be accessed from this project.
```

README.md
```diff
@@ -502,6 +511,114 @@ server:start()
...
+
+### `roles.httpd`
+
+This role can be used only with Tarantool 3. It allows configuring one or more
+HTTP servers. Those servers could be reused by several other roles.
```
oleg-jukovec, line 521:
```suggestion
It allows configuring one or more HTTP servers. Those servers could be reused by several
other roles.
```

README.md
```diff
@@ -502,6 +511,114 @@ server:start()
...
+        name = 'greeting',
+    }, function(tx)
+        return tx:render({text = 'Hello, world!'})
+    end)
+end
+
+M.stop = function()
+    server:delete('greeting')
+end
+
```
oleg-jukovec, line 573: Missed dependency:

https://github.com/tarantool/doc/issues/3883

roles/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+# Install
```
oleg-jukovec, line 1:
```suggestion
# Install.
```

oleg-jukovec (approved):

DerekBum:
Updated everything according to comments.

README.md
oleg-jukovec: Still no dependency.

README.md
oleg-jukovec: Still no dependency.

CHANGELOG.md
DifferentialOrange: > Yes, it is. It uses name, BUT we could delete a route by path too. I don't see any problems.

I'm not sure, but as far as I remember that httpd has that weird thing that allows to set multiple handlers on the same route since route is not a unique identificator (and only one of handlers will be invoked).

DifferentialOrange (approved):

PR #201, opened 2024-11-12
tls: fix error ssl socket support

themilchenko:
Check of ssl socket support occurs always, even if ssl is disabled.

After the patch support of ssl socket checks only if tls is enabled.

### DIFF
--- a/http/server.lua
+++ b/http/server.lua
@@ -1393,7 +1393,7 @@ local function validate_ssl_opts(opts)
         error("ssl_key_file and ssl_cert_file must be set to enable TLS")
     end
 
-    if not sslsocket_supported then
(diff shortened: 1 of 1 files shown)


themilchenko:
> LGTM, but it's better to have a test for the case if it is not too hard (but it looks that it will require a lot of changes in CI and in tests).

Yes, I think that writing a test on this case should done in a separate task.

oleg-jukovec:
> > LGTM, but it's better to have a test for the case if it is not too hard (but it looks that it will require a lot of changes in CI and in tests).
> 
> Yes, I think that writing a test on this case should done in a separate task.

Up to you, but don't forget to create an issue than please.

DerekBum (approved):

PR #203, opened 2024-11-12
roles: add tls support

### LINKED ISSUE #199: role: add SSL support
After the implementation of #35 we could add the TLS support into our role for Tarantool 3:

```yaml
server_name:
  listen: "localhost:3013"
  params:
    ssl_key_file: "path/to/key/file"
    ssl_cert_file: "path/to/key/file"
    ssl_ca_file: "path/to/key/file"
    ssl_ciphers: "cipher1:cipher2"
    ssl_password: "password"
    ssl_password_file: "path/to/ssl/password"
```

themilchenko:
Since we support TLS in this module, we need to add it into roles too.

After the patch TLS params was added. It can be configurated like the following:

```yaml
server_name:
  listen: "localhost:3013"
  ssl_key_file: "path/to/key/file"
  ssl_cert_file: "path/to/key/file"
  ssl_ca_file: "path/to/key/file"
  ssl_ciphers: "cipher1:cipher2"
  ssl_password: "password"
  ssl_password_file: "path/to/ssl/password"
```

Closes #199

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,6 +581,22 @@ end
 return M
 ```
 
+To enable TLS, provide the following params into roles config (for proper work
+it's enough to provide only `ssl_key_file` and `ssl_cert_file`):
+
+```yaml
+roles_cfg:
+  roles.httpd:
+    default:
+        listen: 8081
+        ssl_key_file: "path/to/key/file"
+        ssl_cert_file: "path/to/key/file"
+        ssl_ca_file: "path/to/key/file"
+        ssl_ciphers: "cipher1:cipher2"
+        ssl_password: "password"
+        ssl_password_file: "path/to/ssl/password"
+```
+
 This role accepts a server by name from a config and creates a route to return 
 `Hello, world!` to every request by this route.
 
--- a/README.md
+++ b/README.md
@@ -533,9 +533,9 @@ Example of the configuration:
 roles_cfg:
   roles.httpd:
     default:
-    - listen: 8081
+      listen: 8081
     additional:
-    - listen: '127.0.0.1:8082'
+      listen: '127.0.0.1:8082'
 ```
 
 Server address should be provided either as a URI or as a single port
--- a/roles/httpd.lua
+++ b/roles/httpd.lua
@@ -72,6 +72,19 @@ local function parse_listen(listen)
     return host, port, nil
 end
 
+-- parse_params returns table with set options from config to pass
+-- it into new() function.
+local function parse_params(node)
+    return {
+        ssl_cert_file = node.ssl_cert_file,
+        ssl_key_file = node.ssl_key_file,
+        ssl_password = node.ssl_password,
+        ssl_password_file = node.ssl_password_file,
+        ssl_ca_file = node.ssl_ca_file,
+        ssl_ciphers = node.ssl_ciphers,
+    }
+end
+
 local function apply_http(name, node)
     local host, port, err = parse_listen(node.listen)
     if err ~= nil then
@@ -79,7 +92,8 @@ local function apply_http(name, node)
     end
 
     if servers[name] == nil then
-        local httpd = http_server.new(host, port)
+        local httpd = http_server.new(host, port, parse_params(node))
+
         httpd:start()
         servers[name] = {
             httpd = httpd,
@@ -99,9 +113,14 @@ M.validate = function(conf)
             error("name of the server must be a string")
         end
 
-        local _, _, err = parse_listen(node.listen)
+        local host, port, err = parse_listen(node.listen)
         if err ~= nil then
             error("failed to parse http 'listen' param: " .. err)
+        end
+
+        local ok, err = pcall(http_server.new, host, port, parse_params(node))
+        if not ok then
+            error("failed to parse params in " .. name .. " server: " .. tostring(err))
         end
     end
 end
--- a/test/integration/httpd_role_test.lua
+++ b/test/integration/httpd_role_test.lua
@@ -3,10 +3,15 @@ local treegen = require('luatest.treegen')
 local server = require('luatest.server')
 local fun = require('fun')
 local yaml = require('yaml')
+local fio = require('fio')
+local http_client = require('http.client').new()
+
 
 local helpers = require('test.helpers')
 
-local g = t.group()
+local g = t.group(nil, t.helpers.matrix({use_tls = {true, false}}))
+
+local ssl_data_dir = fio.abspath(fio.pathjoin(helpers.get_testdir_path(), "ssl_data"))
 
 local config = {
     credentials = {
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -581,6 +581,23 @@ end
...
+    roles.httpd:
+        default:
+            listen: 8081
+            params:
+              ssl_key_file: "path/to/key/file"
+              ssl_cert_file: "path/to/key/file"
+              ssl_ca_file: "path/to/key/file"
+              ssl_ciphers: "cipher1:cipher2"
+              ssl_password: "password"
+              ssl_password_file: "path/to/ssl/password"
```
oleg-jukovec, line 598:
```suggestion
roles_cfg:
  roles.httpd:
    default:
      listen: 8081
      params:
        ssl_key_file: "path/to/key/file"
        ssl_cert_file: "path/to/key/file"
        ssl_ca_file: "path/to/key/file"
        ssl_ciphers: "cipher1:cipher2"
        ssl_password: "password"
        ssl_password_file: "path/to/ssl/password"
```

oleg-jukovec, line 598: If you want to be similar to tarantool settings:
```suggestion
            listen:
              uri: 8081
              params:
                ssl_key_file: "path/to/key/file"
                ssl_cert_file: "path/to/key/file"
                ssl_ca_file: "path/to/key/file"
                ssl_ciphers: "cipher1:cipher2"
                ssl_password: "password"
                ssl_password_file: "path/to/ssl/password"
```

oleg-jukovec, line 598: Or we could keep it simple:
```suggestion
        default:
            listen: 8081
            ssl_key_file: "path/to/key/file"
            ssl_cert_file: "path/to/key/file"
            ssl_ca_file: "path/to/key/file"
            ssl_ciphers: "cipher1:cipher2"
            ssl_password: "password"
            ssl_password_file: "path/to/ssl/password"
```

roles/httpd.lua
```diff
@@ -72,14 +72,32 @@ local function parse_listen(listen)
...
+        return {
+            ssl_cert_file = params.ssl_cert_file,
+            ssl_key_file = params.ssl_key_file,
+            ssl_password = params.ssl_password,
+            ssl_password_file = params.ssl_password_file,
+            ssl_ca_file = params.ssl_ca_file,
+            ssl_ciphers = params.ssl_ciphers,
+        }
+    end
+end
```
oleg-jukovec, line 88: We need to validate the params in the role to avoid errors on the `apply` stage (and add unit-tests for the validation too).

oleg-jukovec: If http.server.new() does not have side effects than we could use it for validation in roles at the `validate` call.

README.md
themilchenko: Fixed.

themilchenko: Apply the next suggestion.

themilchenko: Applied this.

roles/httpd.lua
themilchenko: Used new function, added comment to it and unit tests on validation.

test/unit/httpd_role_test.lua
```diff
@@ -122,6 +122,69 @@ local validation_cases = {
...
         err = "failed to parse http 'listen' param: URI query component is not supported",
     },
+    ["ssl_key_file_not_string"] = {
+        cfg = {
+            server = {
+                listen = "localhost:123",
+                ssl_key_file = 123,
+            },
+        },
+        err = "ssl_key_file option must be a string",
```
oleg-jukovec, line 132: Please, add a positive cases too.

test/integration/httpd_role_test.lua
```diff
@@ -3,10 +3,18 @@ local treegen = require('luatest.treegen')
...
+local http_client = require('http.client').new()
+
 
 local helpers = require('test.helpers')
 
-local g = t.group()
+local g = t.group(nil, {
+    { use_tls = false },
+    { use_tls = true },
+})
```
DifferentialOrange, line 15:
```suggestion
local g = t.group(nil, {
    t.helpers.matrix({ use_tls = {false, true} }),
})
```

roles/httpd.lua
```diff
@@ -99,10 +113,15 @@ M.validate = function(conf)
             error("name of the server must be a string")
         end
 
-        local _, _, err = parse_listen(node.listen)
+        local host, port, err = parse_listen(node.listen)
         if err ~= nil then
             error("failed to parse http 'listen' param: " .. err)
         end
+
+        local ok, err = pcall(http_server.new, host, port, parse_params(node))
```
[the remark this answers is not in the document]
themilchenko: new() is just a constructor that only initializes its fields. Why is it dangerous and what do we need to close?

test/integration/httpd_role_test.lua
themilchenko: Isn't matrix uses in case when we want to combine several params? But here we have only one.

Moreover this suggestion isn't working with only one target param: `parameter name should be string, got number`.

test/unit/httpd_role_test.lua
themilchenko: Added.

test/unit/httpd_role_test.lua
```diff
@@ -122,6 +126,78 @@ local validation_cases = {
...
     },
+    ["ssl_ok"] = {
+        cfg = {
+            server = {
+                listen = "localhost:123",
+                ssl_key_file = fio.pathjoin(ssl_data_dir, 'server.key'),
+                ssl_cert_file = fio.pathjoin(ssl_data_dir, 'server.crt')
+            },
+        },
+    },
```
oleg-jukovec, line 137: Please, add a positive test `ssl_full` with all TLS options.

oleg-jukovec (approved):

test/unit/httpd_role_test.lua
themilchenko: Added.

roles/httpd.lua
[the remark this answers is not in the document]
DifferentialOrange: > new() is just a constructor that only initializes its fields

It should be fine then.

test/integration/httpd_role_test.lua
DifferentialOrange: > Isn't matrix uses in case when we want to combine several params? But here we have only one.

Well, it is more useful this way, but even for one parameter it seems to be a bit more readable.

> Moreover this suggestion isn't working with only one matrix param: parameter name should be string, got number.

I think you're made a mistake in array/map mixing somewhere, it works fine

```lua
tarantool> t.helpers.matrix({use_tls = {true, false}})
---
- - use_tls: true
  - use_tls: false
...

tarantool> t.helpers.matrix({use_tls = {true}})
---
- - use_tls: true
...
```

themilchenko: Yes, I'm sorry, just used `matrix` already inside table which caused an error. Fixed.

PR #204, opened 2024-11-14
rockspec: add sslsocket

themilchenko:
Since TLS support was merged, it has been forgotten to add new `sslsocket.lua` file into rockspec. It's impossible to start installed http module without this file.

After the patch `sslsocket.lua` was added into rockspec file.

### DIFF
--- a/http-scm-1.rockspec
+++ b/http-scm-1.rockspec
@@ -28,6 +28,7 @@ build = {
             }
         },
         ['http.server'] = 'http/server.lua',
+        ['http.sslsocket'] = 'http/sslsocket.lua',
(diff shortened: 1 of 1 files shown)


oleg-jukovec (approved):
Please, create an issue to run tests with a code from a rock after build. We already have such test jobs in some modules. It will help to avoid the problem in the future.

themilchenko:
> Please, create an issue to run tests with a code from a rock after build. We already have such test jobs in some modules. It will help to avoid the problem in the future.

Done: https://github.com/tarantool/http/issues/205.

PR #211, opened 2025-06-23
roles: add log_requests option

p0rtale:
Currently, HTTP requests are logged at the default level `log.info`, which can create noise (e.g. for metrics endpoints).
This change allows setting custom log levels (e.g. `log.verbose`) for incoming requests.

Example:
```yaml
roles_cfg:
  roles.httpd:
    default:
      listen: 8081
      log_requests: log.verbose
  roles.metrics-export:
    http:
    - endpoints:
      - path: /metrics
        format: prometheus
```

Closes TNTP-3229

### DIFF
--- a/roles/httpd.lua
+++ b/roles/httpd.lua
@@ -1,5 +1,6 @@
 local checks = require('checks')
 local urilib = require('uri')
+local log = require('log')
 local http_server = require('http.server')
 
 local M = {
@@ -75,7 +76,28 @@ end
 -- parse_params returns table with set options from config to pass
 -- it into new() function.
 local function parse_params(node)
+    local log_requests = node.log_requests
+    if type(log_requests) == "string" then
+        local level = log_requests:lower()
+        if level == "error" then
+            log_requests = log.error
+        elseif level == "warn" then
+            log_requests = log.warn
+        elseif level == "info" then
+            log_requests = log.info
+        elseif level == "verbose" then
+            log_requests = log.verbose
+        elseif level == "debug" then
+            log_requests = log.debug
+        else
+            error("invalid log_requests: " .. log_requests)
+        end
+    elseif log_requests ~= nil then
+        error("log_requests option should be a string")
+    end
+
     return {
+        log_requests = log_requests,
         ssl_cert_file = node.ssl_cert_file,
         ssl_key_file = node.ssl_key_file,
         ssl_password = node.ssl_password,
--- a/test/integration/httpd_role_test.lua
+++ b/test/integration/httpd_role_test.lua
@@ -9,7 +9,20 @@ local http_client = require('http.client').new()
 
 local helpers = require('test.helpers')
 
-local g = t.group(nil, t.helpers.matrix({use_tls = {true, false}}))
+local LOG_LEVEL = {
+    INFO = 5,
+    VERBOSE = 6,
+    DEBUG = 7,
+}
+
+local g = t.group(nil, t.helpers.matrix({
+    use_tls = {true, false},
+    log_level = {
+        LOG_LEVEL.INFO,
+        LOG_LEVEL.VERBOSE,
+        LOG_LEVEL.DEBUG,
+    },
+}))
 
 local ssl_data_dir = fio.abspath(fio.pathjoin(helpers.get_testdir_path(), "ssl_data"))
 
@@ -36,10 +49,16 @@ local config = {
                         ['roles.httpd'] = {
                             default = {
                                 listen = 13000,
+                                log_requests = 'info',
                             },
                             additional = {
                                 listen = 13001,
-                            }
+                                log_requests = 'verbose',
+                            },
+                            additional_debug = {
+                                listen = 13002,
+                                log_requests = 'debug',
+                            },
                         },
                         ['test.mocks.mock_role'] = {
                             {
@@ -49,6 +68,10 @@ local config = {
                                 id = 2,
                                 name = 'additional',
                             },
+                            {
+                                id = 3,
+                                name = 'additional_debug',
+                            }
                         },
                     },
                     instances = {
@@ -79,6 +102,8 @@ g.before_each(function(cg)
     if cg.params.use_tls then
         cfg = tls_config
     end
+
+    cfg.log = {level = cg.params.log_level}
 
     local config_file = treegen.write_file(dir, 'config.yaml',
         yaml.encode(cfg))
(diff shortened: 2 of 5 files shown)


a1div0 (changes requested):
Нужно покрыть минимум одним тестом

http/server.lua
```diff
@@ -1440,7 +1440,7 @@ local exports = {
             cache_templates     = true,
             cache_controllers   = true,
             cache_static        = true,
-            log_requests        = true,
+            log_requests        = false,
```
themilchenko, line 1443: Why do we need to change default behavior?

test/integration/http_server_options_test.lua
```diff
@@ -165,3 +166,25 @@ g.test_idle_timeout_is_unset = function(g)
...
+    g.httpd = helpers.cfgserv({
+        log_requests = log.verbose
+    })
+    g.httpd:start()
+end)
+
+g.test_log_requests_custom = function(g)
+    local httpd = g.httpd
+    t.assert_equals(httpd.options.log_requests, log.verbose)
+end
```
themilchenko, line 190: We already have tests for `log_request` option of `http` module [here](https://github.com/tarantool/http/blob/master/test/integration/http_log_requests_test.lua).

test/unit/httpd_role_test.lua
```diff
@@ -268,3 +269,32 @@ g['test_get_server_bad_type'] = function()
...
+            log_requests = log.verbose,
+        },
+    }
+
+    httpd_role.apply(cfg)
+
+    local result = httpd_role.get_server()
+    t.assert_not_equals(result, nil)
+    t.assert_equals(result.options.log_requests, log.verbose, "Should use custom log level")
+end
```
themilchenko, line 300: I think it can be merged into one test:

* create config with 2 servers with other `log_requests` options;
* check it in for loop (for example). Something like this:

```lua
    for name, body in pairs(cfg) do
        local res = httpd_role.get_server(name)
        t.assert(res)
        t.assert_equals(res.options.log_requests, body.log_requests or false)
    end
```

test/unit/httpd_role_test.lua
```diff
@@ -268,3 +269,32 @@ g['test_get_server_bad_type'] = function()
...
+        [httpd_role.DEFAULT_SERVER_NAME] = {
+            listen = 13002,
+        },
+    }
+
+    httpd_role.apply(cfg)
+
+    local result = httpd_role.get_server()
+    t.assert_not_equals(result, nil)
+    t.assert_equals(result.options.log_requests, false)
```
themilchenko, line 284:
```suggestion
    t.assert(result)
    t.assert_not(result.options.log_requests)
```

test/unit/httpd_role_test.lua
```diff
@@ -268,3 +269,32 @@ g['test_get_server_bad_type'] = function()
...
+        [httpd_role.DEFAULT_SERVER_NAME] = {
+            listen = 13003,
+            log_requests = log.verbose,
+        },
+    }
+
+    httpd_role.apply(cfg)
+
+    local result = httpd_role.get_server()
+    t.assert_not_equals(result, nil)
```
themilchenko, line 298:
```suggestion
    t.assert(result)
```

themilchenko (commented):
Thanks for the patch! Please fix some non-critical comments:

* Amend commits into one if it is about one feature or bug. For example in your case it should be in one commit, because it is all about `log_requests` option.
* Add CHANGELOG entry with info about added functional.

http/server.lua
p0rtale: To reduce log noise by default. Many endpoints (like metrics) don't need info-level logging

themilchenko: > Now the idea is to reduce the importance of logging requests from "info" to "debug"

To be honest, I didn't get a motivation of it. Why does default `info` level not satisfy us? Maybe we can provide non-default level we want manually or even for separate route (for `/metrics` as an example)?

DifferentialOrange: I agree that for now changing the log level for a server (or for a `\metrics` handle) show be the option with the same default as used now. Changing the default is a separate question of breaking backward compatibility that must be processed separately (after some discussion regarding whether or not we should really do it).

a1div0: The main motivation is to avoid clogging the logs at the Info level with this by default:
image

It just interferes with debugging and investigations. Regarding backward compatibility, it is probably worth discussing what can break. As far as I know, in the 3rd version this is only used in metrics.

roles/httpd.lua
```diff
@@ -75,7 +76,22 @@ end
...
 -- it into new() function.
 local function parse_params(node)
+    local log_requests = node.log_requests
+    if type(log_requests) == "string" then
+        local level = log_requests:lower()
+        if level == "info" then
+            log_requests = log.info
+        elseif level == "verbose" then
+            log_requests = log.verbose
+        elseif level == "debug" then
```
themilchenko, line 86: Let's make all levels compatible with Tarantool: https://www.tarantool.io/en/doc/latest/reference/configuration/#confval-log_level

roles/httpd.lua
```diff
@@ -75,7 +76,22 @@ end
...
+            log_requests = log.debug
+        else
+            error("invalid log_requests: " .. log_requests)
```
themilchenko, line 89: Please, add unit test on this validate case.

roles/httpd.lua
p0rtale: Does it make sense to log regular requests at levels above `info` like `warn`, `crit` or `error`? Those levels are typically reserved for exceptional or error situations, not for normal request logging.

test/unit/httpd_role_test.lua
```diff
@@ -209,8 +210,18 @@ local validation_cases = {
...
+            server = {
+                listen = "localhost:123",
+                log_requests = "yes",
+            },
+        },
+        err = "invalid log_requests",
+    }
 }
 
+
```
themilchenko, line 224:
```suggestion
```

roles/httpd.lua
```diff
@@ -75,7 +76,32 @@ end
...
+        end
+    end
+
+    if log_requests ~= nil then
+        if type(log_requests) ~= "function" and type(log_requests) ~= "boolean" then
+            error("log_requests should be a function or a boolean")
+        end
+    end
```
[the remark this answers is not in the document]
p0rtale: Got it, we only allow string types

themilchenko (approved):

test/integration/httpd_role_test.lua
```diff
@@ -146,3 +171,32 @@ g.test_change_server_addr_on_the_run = function(cg)
...
+g.test_log_requests = function(cg)
+    t.skip_if(cg.params.use_tls)
+
+    local function make_request(port)
+        local resp = http_client:get(string.format('http://localhost:%d/ping', port))
+        t.assert_equals(resp.status, 200, 'response not 200')
+    end
+
+    local function assert_should_log(expected)
+        local grep_res = cg.server:grep_log('GET /ping', math.huge)
```
DifferentialOrange, line 184: It's better to use `t.helpers.retrying` rather than infinite grep with `math.huge`.

test/integration/httpd_role_test.lua
```diff
@@ -146,3 +171,32 @@ g.test_change_server_addr_on_the_run = function(cg)
...
+        if expected then
+            t.assert(grep_res)
+        else
+            t.assert_not(grep_res)
+        end
+    end
+
+    local log_level = tonumber(cg.params.log_level)
+
+    make_request(13002)
```
DifferentialOrange, line 194: I think it's better to have something like
```suggestion
    make_request('localhost:13002')
```
or
```suggestion
    make_request({port = 13002})
```
otherwise five digit number is confusing.

DifferentialOrange (commented):
Seems nice, see some minor comments below.

DifferentialOrange (approved):

PR #213, opened 2025-06-23
roles: fix bugs after config reload

### LINKED ISSUE #209: roles: server doesn't change server address after reload
Now `http` role can't change address or port of http server while instance running (with `config:reload()`, for example).

### Reproducer

There is a config with `httpd` role:

```yaml
app:
  file: 'init.lua'

groups:
  group001:
    replicasets:
      replicaset001:
        instances:
          instance001:
            roles: [roles.httpd]
            roles_cfg:
              roles.httpd:
                default:
                  listen: '127.0.0.1:8080'
            iproto:
              listen:
              - uri: '127.0.0.1:3301'
```

Let's start instance:

```bash
$ tt start app
```
Now I want to change a port of default server:

```diff
<                   listen: '127.0.0.1:8080'
---
>                   listen: '127.0.0.1:8081'
```

And apply config with config:reload() command on running instance:

```bash
$ tt connect app
app> require('config'):reload()
```

And check if port was changed:

```bash
$ lsof -i :8080 
COMMAND      PID      USER FD   TYPE  DEVICE SIZE/OFF NODE NAME
tarantool 221085 milchenko 25u  IPv4 1361754      0t0  TCP localhost:http-alt (LISTEN)
$ lsof -i :8081
```

And it's not. Need to fix changing a port of address without restarting an instnace.

### LINKED ISSUE #212: roles: server is not deleted if it is not in use
Now if the server is not used (for example, it was removed from the configuration), it will still be running after the configuration is reloaded.

themilchenko:
This patch-set fixes a several bugs related with config update.

Closes #209 
Closes #212

### DIFF
--- a/roles/httpd.lua
+++ b/roles/httpd.lua
@@ -130,8 +130,20 @@ M.apply = function(conf)
     -- a meaningful error if something goes wrong.
     M.validate(conf)
 
+    local actual_servers = {}
+
     for name, node in pairs(conf or {}) do
+        actual_servers[name] = true
+
         apply_http(name, node)
+    end
+
+    -- Stop server if it is not used anymore.
+    for name, server in pairs(servers) do
+        if actual_servers[name] == nil then
+            server.httpd:stop()
+            servers[name] = nil
+        end
     end
 end
 
--- a/roles/httpd.lua
+++ b/roles/httpd.lua
@@ -99,6 +99,10 @@ local function apply_http(name, node)
             httpd = httpd,
             routes = {},
         }
+    elseif servers[name].host ~= host or servers[name].port ~= port then
+        servers[name].httpd:stop()
+        servers[name].httpd = http_server.new(host, port, parse_params(node))
+        servers[name].httpd:start()
     end
 end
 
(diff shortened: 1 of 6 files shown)


roles/httpd.lua
```diff
@@ -130,9 +134,21 @@ M.apply = function(conf)
...
 
+    local actual_servers = {}
+
     for name, node in pairs(conf or {}) do
+        actual_servers[name] = true
+
         apply_http(name, node)
     end
+
+    -- Stop server if it is not using anymore.
```
DerekBum, line 145:
```suggestion
    -- Stop server if it is not used anymore.
```

test/unit/httpd_role_test.lua
```diff
@@ -268,3 +268,46 @@ g['test_get_server_bad_type'] = function()
     t.assert_not(ok)
     t.assert_str_contains(res, '?string expected, got table')
 end
+
+g.test_stop_unusable_servers = function()
```
DerekBum, line 272: Maybe
```suggestion
g.test_stop_unused_servers = function()
```

roles/httpd.lua
themilchenko: Fixed.

test/unit/httpd_role_test.lua
themilchenko: Corrected, thanks.

Issue #219, opened 2025-10-24
Servers are recreated on every config reload


ita-sammann:
On config reload all HTTP servers are recreated and their routes are deleted even if host and port were not changed in config.
This causes [an issue in metrics-export-role](https://github.com/tarantool/metrics-export-role/issues/38).

The problem is caused by using wrong fields in [roles/httpd.lua](https://github.com/tarantool/http/blob/a932a68d0c3e709eda75d5d624a966eb2a37897a/roles/httpd.lua#L124):
```lua
    elseif servers[name].host ~= host or servers[name].port ~= port then
```
should be
```lua
    elseif servers[name].httpd.host ~= host or servers[name].httpd.port ~= port then
```

Issue #216, opened 2025-08-11
Support not uri options for config reload on the run


themilchenko:
It was possible to change `host` and `port` options in config and apply it after execution `config:reload()` on the server. The same should be done for the ssl and log parameters for the httpd role.

Issue #207, opened 2025-01-20
add support mtls


nwomib:
Please add support mTLS certs

PR #224, opened 2025-11-11
Release 1.9.0

themilchenko:
The release introduces a new `ssl_verify_client` option and changes default behavior with provided `ca_file` param. Also a few bugs were fixed.

### Added

- `ssl_verify_client` option (#207).

### Fixed

- Do not recreate server if it's address and port were not changed (#219).
- Server doesn't change after updating parameters on config reload (#216).


.github/workflows/test.yml
```diff
@@ -12,9 +12,14 @@ jobs:
...
         tarantool: ['1.10', '2.10', '2.11', '3.1', '3.2']
         coveralls: [false]
+        # static-build: [false]
         include:
           - tarantool: '2.11'
             coveralls: true
+            # static-build: false
+          - tarantool: '3.5'
+            coveralls: false
+            # static-build: true
```
DifferentialOrange: Why does it commented?

themilchenko: ~~Like it was before a few old jobs has stucked in pending status, so I returned them to get rid of that~~. No, I forgot to just delete them from branch rules

test/helpers.lua
```diff
@@ -148,4 +148,17 @@ helpers.tcp_connection_exists = function(host, port)
     return ok
 end
 
+local ffi = require('ffi')
+local has_tls_method = pcall(function()
+    return ffi.C.TLS_server_method() ~= nil
+end)
+
+helpers.skip_if_ssl_not_enabled = function()
+    luatest.skip_if(not has_tls_method, 'tarantool does not supports ssl')
```
DifferentialOrange:
```suggestion
    luatest.skip_if(not has_tls_method, 'tarantool does not support ssl')
```

DifferentialOrange (approved):

test/helpers.lua
themilchenko: Fixed, thanks

DifferentialOrange (approved):

PR #229, opened 2026-07-15
test: fix HTTP address reload check

### LINKED ISSUE #228: test: avoid connecting to 0.0.0.0
`test/integration/httpd_role_test::test_change_server_addr_on_the_run` fails in Tarantool CI (see https://github.com/tarantool/tarantool/pull/12918 and failed job [[3]]) after changing the server address from `0.0.0.0:13001` to `localhost:13001`.

The test expects a connection to `0.0.0.0:13001` to fail:

```lua
t.assert_not(helpers.tcp_connection_exists('0.0.0.0', 13001))
```

On Linux, this connection may reach the new listener on `127.0.0.1:13001`, so the assertion is environment-dependent. A similar Linux case [[1]] demonstrates the same behavior. Linux documents `0.0.0.0` as an address for binding to all local interfaces [[2]].

The test should verify `httpd.host` and `httpd.port` after reload and keep the existing request to `localhost:13001` to check that the server is operational.

[1]: https://unix.stackexchange.com/questions/419880/connecting-to-ip-0-0-0-0-succeeds-how-why
[2]: https://man7.org/linux/man-pages/man7/ip.7.html#Special_and_reserved_addresses
[3]: https://github.com/tarantool/tarantool/actions/runs/29399722919/job/87301252904?pr=12918

mandesero:
Connecting to 0.0.0.0 is environment-dependent. On Linux it may reach a listener on 127.0.0.1, so the test can fail even when the server is recreated with the expected address.

This patch verifies the host and port kept by roles.httpd after reload. The existing HTTP request checks that the recreated server is running.

Closes #228


themilchenko (commented):
Thanks for the fix, looks good to me! Could you please rebase on master to fix failed CI?

mandesero:
> Could you please rebase on master to fix failed CI?

Done!

themilchenko (approved):

